PR #199, opened 2018-03-13
Fix auto logout

salomegeo:
@Lyncee59 per out discussion deleted sagas and fixed autoDisconnect middleware. added resetting autodisconnect timer when we reset the field.


plondon:
Nice!

PR #201, opened 2018-03-13
Buy/Sell select partner

prwelber:


packages/blockchain-info-components/src/Form/SelectInput/template.js
```diff
@@ -45,6 +45,7 @@ const Search = styled.input.attrs({ type: 'text' })`
   background-color: ${props => props.theme['white']};
   border: 1px solid ${props => props.theme['gray-2']};
   box-sizing: border-box;
+  padding: 0.5rem 1rem;
```
plondon: Any reason in particular we use rem's here? Not using it anywhere else irc

packages/blockchain-wallet-v4-frontend/src/modals/SfoxExchangeData/styled.js
```diff
@@ -0,0 +1,68 @@
...
+  margin-bottom: 15px;
+`
+
+export const ErrorWrapper = styled.div`
+  margin-top: 5px;
+  a {
+    cursor: pointer;
+    color: ${props => props.theme['brand-secondary']}
+  }
+`
```
plondon: I like the idea of this but maybe want to make it more generic for coinify/unocoin also

PR #213, opened 2018-03-15
Ensure imports all always correctly cased

schnogz:
Adding a plugin that will fail Webpack builds when imports and filenames do not match exactly. This will help avoid issues with TravisCI since its Linux.


merildev:
nice

PR #219, opened 2018-03-20
Upgrade Webpack to v4

schnogz:
Changes:
- Updates Webpack from v2.5 to 4.1.
- Updates all Webpack related dependencies and plugins to latest versions.
- Adds code splitting support.  We now have a JS bundle for application source code and one for 3rd party dependencies.

Results:
- Dev build times: 19% faster
- Prod build times: 56% faster
- File bundle sizes are about 5% smaller

Todo:
- [x] Fully test/validate all builds, development and source code serving scenarios.
- [x] Ensure vendor code file is indeed only 3rd party code.  Since this is a monorepo, `blockchain-wallet-v4` and `blockchain-info-components` source code is getting bundled into vendor file since they are "technically" a dependencies to the `blockchain-wallet-v4-frontend` package


schnogz:
This is fully tested and ready to go. Production bundles are broken but it appears to have been that way for months now.  We will need to investigate that separately.

jtormey:
Amazing, thanks!

PR #234, opened 2018-04-02
Forms

plondon:


packages/blockchain-wallet-v4-frontend/src/modals/SendBitcoin/FirstStep/template.success.js
```diff
@@ -148,33 +159,42 @@ const FirstStep = props => {
...
+          }
+          {
+            feeEditToggled && fee > limits.max && <FeeError size={'12px'} weight={300} color={'error'} onClick={() => props.customFeeHandler(limits.max)}>
+              <FormattedMessage id='modals.sendbitcoin.firststep.feeabovemax' defaultMessage='{max} sat/b or less is recommended' values={{max: limits.max}} />
+            </FeeError>
+          }
         </ColLeft>
         <ColRight>
-          <ComboDisplay coin='BTC'>{totalFee}</ComboDisplay>
+          <ComboDisplay size={'14px'} coin='BTC'>{totalFee}</ComboDisplay>
```
prwelber: This is pretty nitpicky, but I think we can pass the `size` prop as a string. Not sure it needs braces.

plondon: Made the changes and removed other places in the code we use size={string} to be just size=string. Thanks

PR #237, opened 2018-04-03
Add static typing framework Flow

matsjj:
This PR adds static types for the redux/data/bitcoin directory as a first step. 
I also added the integration for CI, it seems like some frontend tests still fail though.
Instructions on how to use it were added to the README. 

This will require some joint effort, ideally everyone contributes to the type system whenever they change a file. We can incrementally increase coverage and should see payoffs relatively soon after.


plondon:
Can you merge master one last time before we merge? Sorry it got stale again

PR #239, opened 2018-04-03
Security Center Design Review

prwelber:


packages/blockchain-wallet-v4/src/types/Wallet.js
```diff
@@ -394,6 +394,12 @@ export const getMnemonic = curry((secondPassword, wallet) => {
   }
 })
 
+export const getMnemonicWithoutPassword = (wallet) => {
```
prwelber: @pernas I needed to access the mnemonic for the security center but the original function was implemented in such a way that if the wallet had a 2nd password, it was required to get the mnemonic (which we don't want for the security center purposes). Let me know if this is OK.

mpfluger: If the wallet has 2nd pw enabled the seedHex is encrypted with the second password. So you can't get the mnemonic without 2nd pw.

prwelber: Ah, thanks Mark. Makes sense. I'll change this.

PR #243, opened 2018-04-04
Add build and runtime support for different environments

schnogz:
**Summary:**
Adds support to build and/or run the application with different configurations. The new `.env` configuration files can be found in `config/env` folder.  Here are two quick examples of the new NPM commands with screenshots below:
 * `yarn build:testnet`
 * `yarn start:testnet`

**_Notes_**
 * All of the new and update NPM scripts are documented in the README. 
 * The configuration files for development and staging are intentionally left out. Developers will need to add those on their own.  
* If Webpack cannot find a desired configuration file (e.g. a developer hasnt created their own `.env.development` config file), it will log a warning and fallback to production config.

**Logging:**
Both the Webpack compilation process and the application will log the apps current configurations to allow for easier debugging and troubleshooting. See screenshots below:

**Webpack Log**
webpack-logs

**Application Console Log**
screen shot 2018-04-05 at 12 33 19 pm


schnogz:
Closing this PR as the approach has changed and this problem is now fixed via PR #251

PR #245, opened 2018-04-04
wallet-options

plondon:
Trying to remove hardcoded values that we should load from wallet-options. This fetches wallet-options at `Public/` from '/Resources/wallet-options.json' then saves the url's for core on `global.`.


packages/blockchain-wallet-v4-frontend/webpack.config.js
```diff
@@ -4,6 +4,7 @@ const CleanWebpackPlugin = require('clean-webpack-plugin')
 const CaseSensitivePathsPlugin = require('case-sensitive-paths-webpack-plugin')
 const UglifyJSPlugin = require('uglifyjs-webpack-plugin')
 const BundleAnalyzerPlugin = require('webpack-bundle-analyzer').BundleAnalyzerPlugin
+const walletOptions = require('../Resources/wallet-options.json')
```
plondon: @schnogz we could load the 4 wallet options here for each env then use some logic in the before function inside devServer to determine which walletOptions to return.

schnogz: Yeah that would work. I can lift some of the code from my environment PR into here to get that working.

plondon:
Closing in favor of https://github.com/blockchain/blockchain-wallet-v4-frontend/pull/243.

PR #251, opened 2018-04-09
Refactoring - added mandatory prefetch for wallet-options

Lyncee59:
- We prefetch the wallet-options before starting the sagaMiddleware
- We pass down the api, socket and options to the rootSaga and its child sagas.
(Note that options will be reduced directly as we just fetched them)

-> We need to also pay attention to coinifyService and sfoxService that I removed from the rootSaga parameters


plondon:
Moved necessary functions from coinifyService and sfoxService to core: https://github.com/blockchain/blockchain-wallet-v4-frontend/pull/252/

schnogz:
Here is a list of work I found with some digging....

- [ ] `blockchain-wallet-v4/src/network/socket.js` has hardcoded URIs
- [ ] `blockchain-wallet-v4/src/api/index.js` has hardcoded URIs
- [ ] `blockchain-wallet-v4/src/api/options/index.js` references a local copy of wallet-options.json
- [ ] `frontend/src/config` folder has hardcoded URIs and settings. 
- [X] Move sfoxService and coinifyService to core

PR #255, opened 2018-04-10
Addresses page related bugs

sixtedemaupeou:


packages/blockchain-wallet-v4-frontend/src/scenes/Home/BalancesChart/template.success.js
```diff
@@ -86,10 +86,10 @@ const BalancesChart = (props) => {
             <FormattedMessage id='scenes.home.balanceschart.bch' defaultMessage='Bitcoin Cash' />
           </Text>
           <CoinBalance onClick={handleCoinDisplay}>
-            <SwitchableDisplay coin='BCH' size='14px' weight={200}>{bchBalance}</SwitchableDisplay>
+            <SwitchableDisplay coin='BCH' cursor='pointer' size='14px' weight={200}>{bchBalance}</SwitchableDisplay>
```
plondon: Cant we just do cursor: pointer in one or two places (SwitchableDisplay, FiatDisplay, etc) instead of inline?

packages/blockchain-wallet-v4/src/types/HDWallet.js
```diff
@@ -66,7 +66,7 @@ export const deriveAccountNodeAtIndex = (seedHex, index, network) => {
 
 export const generateAccount = curry((index, label, seedHex) => {
   let node = deriveAccountNodeAtIndex(seedHex, index, Bitcoin.networks.bitcoin)
-  return HDAccount.fromJS(HDAccount.js(`${label} ${index + 1}`, node))
+  return HDAccount.fromJS(HDAccount.js(label, node))
```
plondon: What happens on wallet recovery? Is there a label defined?

packages/blockchain-wallet-v4-frontend/src/scenes/Home/BalancesChart/template.success.js
sixtedemaupeou: I thought about that but I don't think all SwitchableDisplays and FiatDisplays should have cursor: pointer

sixtedemaupeou: Yes you're right, we should do that but only for SwitchableDisplay

packages/blockchain-wallet-v4/src/types/HDWallet.js
sixtedemaupeou: On wallet recovery the labels are defined in the js method of types/HDWallet.js with ```HDAccount.js(`${label} ${i + 1}`, node(i), xpub)``` not in generateAccount

PR #262, opened 2018-04-11
RNG

jtormey:
Ports rng module from v3, mixes local and server entropy.

### DIFF
--- a/packages/blockchain-wallet-v4/package.json
+++ b/packages/blockchain-wallet-v4/package.json
@@ -34,15 +34,6 @@
       "<rootDir>/lib"
     ]
   },
-  "standard": {
-    "globals": [
-      "jest",
-      "describe",
-      "it",
-      "expect",
-      "beforeEach"
-    ]
-  },
   "dependencies": {
     "bech32": "^1.1.3",
     "big-rational": "^0.10.7",
--- a/packages/blockchain-wallet-v4/package.json
+++ b/packages/blockchain-wallet-v4/package.json
@@ -34,6 +34,15 @@
       "<rootDir>/lib"
     ]
   },
+  "standard": {
+    "globals": [
+      "jest",
+      "describe",
+      "it",
+      "expect",
+      "beforeEach"
+    ]
+  },
   "dependencies": {
     "bech32": "^1.1.3",
     "big-rational": "^0.10.7",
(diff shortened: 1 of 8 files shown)


packages/blockchain-wallet-v4/src/walletCrypto/rng.js
```diff
@@ -0,0 +1,131 @@
...
+
+// getServerEntropy :: Int -> Promise (Either Buffer Error) Error
+export const getServerEntropy = (nBytes = DEFAULT_BYTES, api) => {
+  if (!isPositiveInteger(nBytes)) {
+    return Promise.resolve(Either.Left(
+      new Error('Must provide a positive integer to getServerEntropy')
+    ))
+  }
+
+  return api.getRandomBytes(nBytes, FORMAT).then(responseText => {
```
mpfluger: Are non 200 error codes handled further upstream and result in an error that's caught and re-thrown in the catch below?

mpfluger:
@akly and @kristovatlas please review this as well. Most of the code is from the V3 wallet with some minor refactoring: https://github.com/blockchain/My-Wallet-V3/blob/master/src/rng.js.

packages/blockchain-wallet-v4/src/walletCrypto/rng.js
jtormey: A failed request will cause the promise returned by `createRng` to reject, so it's up to the caller to handle it. The code in this block only runs if the call randombytes was successful.

mpfluger: OK, thanks!

packages/blockchain-wallet-v4/package.json
```diff
@@ -34,6 +34,15 @@
       "<rootDir>/lib"
     ]
   },
+  "standard": {
```
schnogz: Shouldn't need this since we use ESLint and utilize Jest env support in the root `.eslintrc` file.

schnogz: Also PR #263 adds support for ECMAScript 8 which once merged will fix your lint "violations"

jtormey: Thanks! Will fix

packages/blockchain-wallet-v4/src/utils/checks.js
```diff
@@ -0,0 +1,13 @@
+import { is } from 'ramda'
+
+/* Number */
+export const isNumber = is(Number)
```
pernas: for the payment branch we have those things defined in the payment file. We will need to remember to use these ones too :D

jtormey: Yes that's a good idea

PR #264, opened 2018-04-11
Prod Builds

plondon:


packages/blockchain-wallet-v4-frontend/src/index.prod.js
```diff
@@ -28,4 +28,5 @@ configureStore().then(x => {
   renderApp(App, x.store, x.history)
 }).catch(e => {
   renderError()
+  console.log(e)
```
plondon: Think it might be nice to have this for worst case scenario

PR #275, opened 2018-04-13
CI Docker Setup

schnogz:
First crack at it....


server.js
```diff
@@ -0,0 +1,87 @@
...
+const path = require('path')
+
+// store env configs
+const port = process.env.PORT || 8080
+const rootURL = process.env.ROOT_URL
+const webSocketURL = process.env.WEB_SOCKET_URL
+const apiDomain = process.env.API_DOMAIN
+const iSignThisDomain = process.env.I_SIGN_THIS_DOMAIN
+const isProduction = rootURL === 'https://blockchain.info'
+
```
plondon: What do you think of adding a process.env.ENV=production instead, so we don't have to check against .info?

Dockerfile
```diff
@@ -0,0 +1,28 @@
+FROM docker-registry.service.consul:5000/blockchain_javascript@sha256:9ae5d67167dc8d9ccc0bae57b3ba7a07f241ad59fa91a85d3eec9d2594b1c471
+#FROM node:9.11.1
+
+# environment config fallbacks that can be overriden on startup
+# e.g. docker run -it -p 8080:8080 -e ROOT_URL="whatever"[image-name]
+ENV ROOT_URL='https://blockchain.info'
+ENV WEB_SOCKET_URL='wss://ws.blockchain.info/inv'
+ENV API_DOMAIN='https://api.blockchain.info'
+ENV I_SIGN_THIS_DOMAIN='https://stage-verify.isignthis.com'
```
plondon: We don't need IST_DOMAIN here because it's prod ? https://verify.isignthis.com : https://stage-verify.isignthis.com

We do need the wallet-helper domain however

server.js
schnogz: 👍

Dockerfile
schnogz: yeah i will make a list of build secrets that will need to be added/modified

PR #281, opened 2018-04-16
SFOX CYO

jtormey:


jtormey:
Hmm doesn't seem eslint can handle babel stage-2 syntax.

PR #283, opened 2018-04-16
Add missing bch accounts on fetch metadata

sixtedemaupeou:


plondon:
Is it also possible to call this after BTC new account creation?

PR #286, opened 2018-04-16
Fixes issues with running production code locally

schnogz:
This will allow developers to run production code locally using the exact production setup.  This means that HMR is disabled, code is optimized and bundled and served from the production Express server instead of the Webpack dev-sever.

Other Changes:
- Removed excess logging from Dockerfile
- Renamed environment config property keys to more closely resemble the CI
- Renamed `isProd` type build flags to `isCiBuild` to avoid confusion
- Update readme


plondon:
Nice. Glad we can also use wallet-options with isLocal flag.

PR #291, opened 2018-04-17
Export eth private keys

sixtedemaupeou:
- [X] Export main account private key
- [x] Export legacy account private key


packages/blockchain-wallet-v4-frontend/src/modals/ShowEthPrivateKey/index.js
```diff
@@ -0,0 +1,46 @@
...
+        {...this.props}
+        step={step}
+        onContinue={nextStep}
+      />
+    )
+  }
+}
+
+const mapStateToProps = (state, ownProps) => {
+  const ethKvStoreSelectors = selectors.core.kvStore.ethereum
```
Lyncee59: This should live in a separate file (selectors.js), to make it easier to unit test

plondon:
@sixtedemaupeou  It looks like the private key for main/legacy is different, but the public key is the same.

plondon:
One more thing, if I don't have a legacy eth account I should only see 'Export Private Key'

PR #296, opened 2018-04-18
Cache chart in localStorage in case of future failure

sixtedemaupeou:


packages/blockchain-wallet-v4-frontend/src/scenes/Login/index.js
```diff
@@ -16,6 +16,8 @@ class LoginContainer extends React.Component {
   onSubmit (event) {
     event.preventDefault()
     const { guid, password, code } = this.props
+    localStorage.setItem('ls.guid', JSON.stringify(guid))
```
plondon: I think we're doing this a bit too soon here, and should instead save it after a successful login attempt. We should also save the session token. See: https://github.com/blockchain/My-Wallet-V3-Frontend/blob/4e892c2363b4bd3b3c4d8c7aed053b7586bc4b93/assets/js/services/wallet.service.js#L172-L181

packages/blockchain-wallet-v4-frontend/src/scenes/Home/PriceIndexSeries/Chart/index.js
```diff
@@ -41,14 +41,28 @@ class Chart extends React.Component {
     const { start, interval } = selectPriceIndexSeriesOptions(coin, timeframe)
```
plondon: I see ('ls.chart') item in my localStorage, but doesn't look we're using that to load the latest 'settings/config' of the chart.

sixtedemaupeou: We are using it to load the config of the chart if the selector getPriceIndexSeries returns a Remote.Failure. 
`Failure: (message) => {
        const value = JSON.parse(window.localStorage.getItem('ls.chart-data'))
        return isNil(value)
          ? (<Error>{message}</Error>)
          : (<Success
            coin={coin}
            timeframe={timeframe}
            config={configure(start, interval, currency, value)}
            selectCoin={this.selectCoin}
            selectTimeframe={this.selectTimeframe}
          />)
      },`

PR #310, opened 2018-04-20
Whats new styling

TheLeoB:


TheLeoB:
I don't think the remaining indentation errors need to be fixed. Is there a way to bypass?

plondon:
No way to bypass, I recommend installing a linter for your text editor. Let me know if I can help.

PR #311, opened 2018-04-20
Misc fixes

plondon:


packages/blockchain-info-components/src/Text/TextGroup.js
```diff
@@ -5,13 +5,13 @@ import styled from 'styled-components'
...
   ${props => props.nowrap ? '& > * { white-space: nowrap }' : '& > * { white-space: normal }'};
-  ${props => props.inline ? '& > * { display: inline; margin-right: 5px!important; }' : '& > * { display: block; margin-bottom: 10px; }'};
+  ${props => props.inline ? '& > * { display: inline; }' : '& > * { display: block; margin-bottom: 10px; }'};
 `
 const TextGroup = props => {
   const { children, ...rest } = props
 
   return (
-    <BaseTextGroup {...rest}>{children}</BaseTextGroup>
+    <BaseTextGroup {...rest}>{children}{' '}</BaseTextGroup>
```
sixtedemaupeou: What's the point of this space? It looks like a hack to fix a bug. Also, with the space you added in Text components it means there will be 2 spaces at the end of any TextGroup ending with a Text.

plondon: It applies a `&nbsp;` after the child. We could solve the issue with CSS instead and apply `&:after { content: '&nbsp;' }` to every child except the last child.

PR #332, opened 2018-04-27
Wallet & Addresses Preferences Page Updates

schnogz:
- Adds most of the needed functionality for used addresses support which is currently hidden from UI
- Adds search/filter capability to unused addresses page
- Fixes bugs related to archived addresses not being labeled correctly
- Adds plus icon to iconset
- Updates wallets, imported addresses and archived addresses tables to use consistent banners, column headers and spacing.


packages/blockchain-wallet-v4-frontend/src/scenes/Settings/Addresses/Btc/Wallets/selectors.js
```diff
@@ -1,7 +1,18 @@
...
+  const wallets = map(x => ({
+    label: x.label,
+    index: x.index,
+    archived: x.archived,
+    default: defaultId === x.index,
+    balance: x.info.final_balance,
+    xpub: x.xpub
+  }))
+
+  return selectors.core.common.bitcoin.getActiveHDAccounts(state).map(wallets)
```
plondon: Is there a getHDAccounts selector instead? Noticed that if I archive an account there is no way to unarchive it.

PR #333, opened 2018-04-27
feat(Icons): add shapeshift-switch

plondon:


plondon:
I messed up and didn't add the icomoon font files. Will add them after andrew's settings branch is merged.

PR #342, opened 2018-04-30
Save BTC transaction notes

schnogz:
Also fixes a few other small copy and spacing issues.


plondon:
We need to do one more thing here, if a tx has a description show that in the tx feed, if there is also a label, prefer that

plondon:
This also resolves this TODO: https://github.com/blockchain/blockchain-wallet-v4-frontend/blob/master/packages/blockchain-wallet-v4/src/redux/payment/btc/sagas.js#L27-L28

PR #347, opened 2018-05-01
Local storage cache

sixtedemaupeou:


plondon:
I see this when logging in:
screen shot 2018-05-03 at 9 29 05 am

PR #364, opened 2018-05-03
fix - performance issues on activity list

Lyncee59:
- [x] Spot checked


packages/blockchain-wallet-v4-frontend/src/scenes/Home/ActivityList/index.js
```diff
@@ -32,15 +24,11 @@ class ActivityListContainer extends React.PureComponent {
 }
 
 const mapStateToProps = (state, ownProps) => ({
-  data: getData(state, 8),
-  ethContext: getEthereumContext(state)
+  data: getData(state, 8)
```
plondon: I don't think 8 is used here anymore?

PR #375, opened 2018-05-04
Feature/performance btc transactions

Lyncee59:
- [x] Spot checked
- Improvement bitcoin transaction page, using reselect
- Added opt in button to show the value when received/sent to avoid performing X Ajax queries prior to rendering.
- Refactoring bitcoin transactions page using saga instead of react lifecycles.


packages/blockchain-wallet-v4-frontend/src/components/TransactionListItem/FiatAtTime/ValueWhenReceived/selectors.js
```diff
@@ -0,0 +1,9 @@
+import { createSelector } from 'reselect'
+import { curry } from 'ramda'
+import { Remote } from 'blockchain-wallet-v4/src'
+import { selectors } from 'data'
+
+export const getData = curry((hash, currency, currencySymbol, state) => createSelector(
+  [state => selectors.core.data.bitcoin.getFiatAtTime(hash, currency)(state)],
+  (fiatR) => (fiatR || Remote.NotAsked).map(x => `${x} ${currencySymbol}`)
```
sixtedemaupeou: Use FiatDisplay component or something like that

Lyncee59: FiatDisplay converts a Coin amount to a Fiat amount, so I can't use that. I created a new function to format a fiat amount in the Exchange service => formatFiat(value, currency)

PR #402, opened 2018-05-09
Fix Yubikey 2FA issue

schnogz:
## Description
- Fixes issue where Yubikey codes were being uppercased when they should not have been resulting in login failures.
- Makes the Yubikey input a password box on login page

## Change Type
Bug Fix

## Testing Steps
Create account with 2FA Yubikey and attempt to login

## PR Creator Checklist
- [x] Code compiles correctly

## PR Reviewer Checklist

### DIFF
--- a/packages/blockchain-wallet-v4-frontend/src/scenes/Login/template.js
+++ b/packages/blockchain-wallet-v4-frontend/src/scenes/Login/template.js
@@ -11,6 +11,7 @@ import { Form, FormError, FormGroup, FormItem, FormLabel, PasswordBox, TextBox }
 import Modals from 'modals'
 import MobileLogin from 'modals/MobileLogin'
 
+const Fragment = React.Fragment
 const Wrapper = styled.div`
   width: 100%;
   padding: 35px;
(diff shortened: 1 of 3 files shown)


packages/blockchain-wallet-v4-frontend/src/scenes/Login/template.js
```diff
@@ -128,12 +129,30 @@ const Login = (props) => {
         { authType > 0 &&
           <FormGroup>
             <FormItem>
-              <FormLabel for='code'>
-                { authType === 1 && <FormattedMessage id='scenes.login.yubikey' defaultMessage='Yubikey' /> }
-                { authType === 4 && <FormattedMessage id='scenes.login.google' defaultMessage='Authenticator App Code' /> }
-                { authType === 5 && <FormattedMessage id='scenes.login.mobile' defaultMessage='SMS Code' /> }
-              </FormLabel>
-              <Field name='code' validate={[required]} component={TextBox} borderColor={twoFactorError ? 'invalid' : undefined} />
```
plondon: Out of curiosity does `authType === 1 ? PasswordBox : TextBox` work?

plondon: If not you could create a function probably that does return the correct component instead of 3 cases

schnogz: fixed

PR #443, opened 2018-05-24
Ethereum 

salomegeo:
## Description
display warning when trying to send ether and there's unconfirmed transaction.
## Change Type
- Bug Fix

## PR Creator Checklist
- [x] Code compiles correctly

## PR Reviewer Checklist
- [x] Change validated and application spot checked
- [x] Code styles and best practices met
- [x] Code is readable and commented when necessary


plondon:
At first I thought it was kind of weird that unconfirmedTransaction was inside the payment, but then realized how valuable that will be on things like shapeshift/exchange when we need to wait on these components as well. @Lyncee59

plondon:
@salomegeo looks good for the most part! We need to continue using last_tx and last_tx_timestamp from v3 metadata, instead of putting both in a last_tx object (you're way is probably better but all platforms use what we had in v3, so we must continue to support them). I've also disabled `To:` field. Amount is WIP.

packages/blockchain-wallet-v4/src/redux/payment/eth/sagas.js
```diff
@@ -42,8 +42,30 @@ export default ({ api }) => {
         const fees = yield call(api.getEthereumFee)
         const gasPrice = prop('regular', fees)
         const gasLimit = prop('gasLimit', fees)
-        const fee = calculateFee(gasPrice, gasLimit)
-        return makePayment(merge(p, { fees, fee }))
+        const fee = calculateFee(0.00000000001, gasLimit)
```
salomegeo: This probably shouldnt' be committed

salomegeo:
agh, I thought there would be like a three line change, that I just didn't know how to do :(

PR #444, opened 2018-05-24
Feature/misc bugs 2

Lyncee59:
## Description
Different bugs fixes on exchange

## Change Type
- Bug fixes related to exchange

## Testing Steps

## PR Creator Checklist
- [x] Code compiles correctly
- [ ] All unit tests pass (verified via `yarn test`)
- [ ] Updated `README.md` and other documentation (if necessary)

## PR Reviewer Checklist


packages/blockchain-info-components/src/Form/NumberInput.js
```diff
@@ -1,27 +1,31 @@
...
   font-weight: 300;
   color: ${props => props.theme['gray-5']};
-  background-color: ${props => props.theme['white']};
+  background-color: ${props => props.disabled ? props.theme['gray-1'] : props.theme['white']};
   font-family: 'Montserrat', Helvetica, sans-serif;
   background-image: none;
   outline-width: 0;
   user-select: text;
   border: 1px solid  ${props => props.theme[props.borderColor]};
+  cursor: ${props => props.disabled ? 'not-allowed' : 'initial'};
```
plondon: Why not just use the browsers default disabled functionality by setting the input to `disabled="true"`

packages/blockchain-wallet-v4-frontend/src/data/components/exchange/sagas.utils.js
```diff
@@ -116,8 +118,8 @@ export default ({ api, coreSagas }) => {
         if (isUndefinedOrEqualsToZero(sourceAmount)) return defaultResult
         const quotation = yield call(api.createQuote, sourceAmount, pair, true)
         const targetAmount = path(['success', 'withdrawalAmount'], quotation) || 0
-        const sourceFiat = convertCoinToFiat(sourceAmount, sourceCoin, sourceCoin, 'USD', sourceRates).value
-        const targetFiat = convertCoinToFiat(targetAmount, targetCoin, targetCoin, 'USD', targetRates).value
+        const sourceFiat = convertCoinToFiat(sourceAmount, sourceCoin, sourceCoin, currency, sourceRates).value
+        const targetFiat = convertCoinToFiat(targetAmount, targetCoin, targetCoin, currency, targetRates).value
         return { sourceAmount, sourceFiat, targetAmount, targetFiat }
```
plondon: In v3 the currency is USD if the user is in the US, EUR otherwise, not sure we should change this. Also there is a limit set from wallet-options of 500 (USD if user is US, EUR otherwise)

Lyncee59: I never heard about the 500 limits ? does it mean we can't validate the trade if fiat amount > 500 ?

packages/blockchain-info-components/src/Form/NumberInput.js
Lyncee59: Actually, I just reused what was already created on TextInput. I can't remember the reason why I did not use the html attribute indeed, will investigate

Lyncee59: Ok well I removed it. Now the input is disabled but just show the default cursor | when you are on hover.

packages/blockchain-wallet-v4-frontend/src/data/components/exchange/sagas.utils.js
plondon: Sorry I should have brought it up before
1. the form should be invalid if amount > 500
2. fiat = country_code === US ? USD : EUR

PR #461, opened 2018-05-29
websocket inconsistencies

salomegeo:
## Description
fetch transactions after send, we might not receive message from soket.
display receive notifications only when received
853, 702, 868

## Change Type
- Bug Fix

## PR Creator Checklist
- [x] Code compiles correctly

## PR Reviewer Checklist
- [x] Change validated and application spot checked
- [x] Code styles and best practices met
- [x] Code is readable and commented when necessary


packages/blockchain-wallet-v4/src/redux/webSocket/bch/sagas.js
```diff
@@ -19,9 +20,14 @@ export default ({ api, bchSocket }) => {
...
-        const walletContext = yield select(walletSelectors.getWalletContext)
-        yield put(bchActions.fetchData(walletContext))
         yield put(bchActions.fetchTransactions('', true))
+        const transactions = yield take(bchAT.FETCH_BCH_TRANSACTIONS_SUCCESS)
+        for (let i in transactions.payload.transactions) {
+          const tx = transactions.payload.transactions[i]
+          if (tx.hash === message.x.hash && tx.result > 0) {
+            yield put(A.webSocket.bitcoin.paymentReceived('You\'ve just received a bitcoin payment.'))
+          }
+        }
```
plondon: v3 does something a bit cheaper and only checks the first tx and I don't remember issues with inconsistency. What do you think?

```
      let tx = wallet.my.wallet.txList.transactions()[0];
      if (tx.result > 0 && tx.txType === 'received') {
        Alerts.displayReceivedBitcoin('JUST_RECEIVED_BITCOIN');
      }```

salomegeo:
Ok make a small change, mostly this will only check the first transaction.

Inconsistencies are that there is no clear pattern(let alone guarantees) how websockets will behave for sent transactions. So both for btc and bch:
1. if we don't receive utx notification for sent transactions we fetch transactions after send(this updates the balance too.)
2. if we do recieve utx mesage for sent transactions we don't display the notification.

PR #477, opened 2018-06-01
fiat conversion

salomegeo:
## Description
use Number.toFixed instead of BigRational.toDecimal, because toDecimal always rounds it down
819

## Change Type
- Bug Fix

## PR Creator Checklist
- [ ] Code compiles correctly
- [ ] All unit tests pass (verified via `yarn test`)

## PR Reviewer Checklist

### DIFF
--- a/packages/blockchain-wallet-v4/src/exchange/currency/index.js
+++ b/packages/blockchain-wallet-v4/src/exchange/currency/index.js
@@ -13,7 +13,7 @@ export class Currency extends Type {
   toUnit (unit) {
     if (unit && unit.currency === this.currency.code) {
       return Maybe.Just({
-        value: this.value.multiply(BigRational(unit.rate).reciprocate()).toDecimal(unit.decimal_digits),
+        value: Number(this.value.multiply(BigRational(unit.rate).reciprocate())).toFixed(unit.decimal_digits),
         unit: unit
       })
     } else { return Maybe.Nothing() }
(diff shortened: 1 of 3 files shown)


packages/blockchain-wallet-v4/src/exchange/currency/index.js
```diff
@@ -13,7 +13,7 @@ export class Currency extends Type {
   toUnit (unit) {
     if (unit && unit.currency === this.currency.code) {
       return Maybe.Just({
-        value: this.value.multiply(BigRational(unit.rate).reciprocate()).toDecimal(unit.decimal_digits),
+        value: Number(this.value.multiply(BigRational(unit.rate).reciprocate())).toFixed(unit.decimal_digits),
```
plondon: `Number(5.00.toFixed(2))`, is this what you want?

salomegeo: Number(5.234).toFixed(2)

plondon: But trailing 0's will still be mitigated by formatFiat, which we're not getting rid of?

salomegeo: This shouldn't affect trailing 0's at all.

I'm not completely sure about trailing 0's in general, where do we want and where do we not want them. If it's consistent we can move that in here too, but if not it should be handled case-by-case bases, which I think is what you did in your pull request.

PR #486, opened 2018-06-05
feat - updated DateInput to include a calendar to pick the date

Lyncee59:
## Description
Implementation of calendar using react-datetime library

## Change Type
- Feature

## Testing Steps

## PR Creator Checklist
- [x] Code compiles correctly

## PR Reviewer Checklist


Lyncee59:
I had to use the CSS provided with the component, however to make it compatible with theme, I imported all this CSS directly in the top styled-component. It's definitely not the idea solution, as further version of react-datetime may need an update of this CSS, but I think we should at some point take the time to create our inhouse calendar, so its gonna be good for now

PR #519, opened 2018-06-08
Release/v4.0.0 alpha11

plondon:


packages/blockchain-wallet-v4-frontend/src/scenes/SecurityCenter/EmailAddress/index.js
```diff
@@ -47,8 +50,8 @@ class EmailAddressContainer extends React.PureComponent {
   }
 
   handleChangeEmailView () {
+    const { email } = this.props.data.getOrElse({})
     this.props.updateUI({ changeEmailToggled: true })
-    const email = this.props.data.data.email
     this.props.formActions.change('securityEmailAddress', 'changeEmail', email)
```
sixtedemaupeou: Should you still do that if email is undefined?

plondon: When is email undefined?

sixtedemaupeou: Whenever this.props.data is not a Remote.Success

sixtedemaupeou: I guess that happens if we fail to fetch the settings or if they haven't been fetched yet

plondon: I think settings are fetched when user logs in.

If they reach security center and settings are not fetched the worst I think will happen is the field will be empty, instead of prefilled with your email by default.

Anyway, think this is better than data.data, which would break HMR

PR #539, opened 2018-06-11
Improve countdown timer for SFOX quotes

prwelber:
## Description
840 and a small Coinify change

## Change Type
- Bug Fix

## Testing Steps

## PR Creator Checklist
- [x] Code compiles correctly

## PR Reviewer Checklist

### DIFF
--- a/packages/blockchain-wallet-v4-frontend/src/components/Form/CountdownTimer/index.js
+++ b/packages/blockchain-wallet-v4-frontend/src/components/Form/CountdownTimer/index.js
@@ -7,8 +7,13 @@ class CountdownTimerContainer extends React.PureComponent {
   constructor (props) {
     super(props)
     this.interval = undefined
-    const { expiryDate } = this.props
-    this.state = { expiration: moment(expiryDate), elapsed: moment.duration(moment(expiryDate).diff(moment())) }
+    const { expiryDate, createdDate } = this.props
+    let startTime = createdDate || undefined
+    this.state = {
+      start: moment(startTime),
+      expiration: moment(expiryDate),
+      elapsed: moment.duration(moment(expiryDate).diff(moment(startTime)))
+    }
     this.tick = this.tick.bind(this)
   }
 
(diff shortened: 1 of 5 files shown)


packages/blockchain-wallet-v4-frontend/src/components/Form/CountdownTimer/index.js
```diff
@@ -22,13 +27,14 @@ class CountdownTimerContainer extends React.PureComponent {
...
-    const elapsed = moment.duration(moment(this.state.expiration).diff(moment()))
+    const elapsed = moment.duration(moment(this.state.expiration).diff(moment(this.state.start)))
+
     if (this.state.elapsed.as('seconds') < 1) {
       // If we reach the end of the timer, we execute the expiry callback
       if (handleExpiry) { handleExpiry() }
     } else {
       // We increment the time elapsed
-      this.setState({ elapsed })
+      this.setState({ elapsed, start: moment(this.state.start).add(1, 'seconds') })
```
plondon: Why do we have to add 1 second to start time? Shouldn't start time be constant?

prwelber: With each "tick" we need to increment from the start time, otherwise the timer will be getting the `diff` between two constant times and it will always be the same. The idea of this approach is to take the internal clock of the computer/browser out of the equation since it can vary.

prwelber: That might not have been clear. The start time is constant of course, but we need a time to move so we can measure it against expiration time. If no start time is given, the computer's clock will be on the move and we'll get the diff from that. Since we don't want to look at that for certain situations, we'll have to set the start time and then increment and keep track in the component's state. Hope that helps to explain.

plondon: I think the naming confused me. `elapsed` is really `remaining`

PR #529, opened 2018-06-11
remove session token when remember 2fa is disabled

salomegeo:
pass true instead of 1 to api call
fixes 665 and more.

## Change Type
- Bug Fix

## PR Creator Checklist
- [ ] Code compiles correctly
- [ ] No lint issues exist (verified via `yarn lint`)
- [ ] All unit tests pass (verified via `yarn test`)
- [ ] Updated `README.md` and other documentation (if necessary)

## PR Reviewer Checklist
- [ ] Change validated and application spot checked
- [ ] Code styles and best practices met
- [ ] Code is readable and commented when necessary

PR #542, opened 2018-06-11
feat(Browser deauthorization support)

schnogz:
## Description
Adds ability for users to deauthorize their browsers which will require them to verify email and 2FA upon next login.  A 10 second timer is started when the `/logout` route is hit. If the user doesn't make a decision in that time, they will be redirected to `/login`.

screen shot 2018-06-11 at 6 51 57 pm

## Change Type
Feature

## Testing Steps
Login using a 2FA enabled wallet and upon logout, select deauthorize browser and verify 2FA will be again required.

## PR Creator Checklist
- [x] Code compiles correctly


plondon:
Something I didn't realize but salome's PR (https://github.com/blockchain/blockchain-wallet-v4-frontend/pull/529/files#diff-80bc26c1898f9919cd89112ea9078d4aR4) reminded me, do we also want to remove the session from local storage? Not sure why since de-authorize still requires me to get a new one though...

PR #550, opened 2018-06-12
Coinify - Initialize Checkout

prwelber:
## Description
1001

## Change Type
- Bug Fix

## Testing Steps

## Code Checklist

### DIFF
--- a/packages/blockchain-wallet-v4-frontend/src/data/modules/coinify/sagas.js
+++ b/packages/blockchain-wallet-v4-frontend/src/data/modules/coinify/sagas.js
@@ -298,7 +298,6 @@ export default ({ coreSagas }) => {
     const trade = yield select(selectors.core.data.coinify.getTrade)
 
     if (path(['type'], head(modals)) === 'CoinifyExchangeData') {
-      yield put(A.initializeCheckoutForm('buy'))
       yield put(actions.modals.closeAllModals())
     } else if (trade.data.state === 'awaiting_transfer_in') {
       yield put(actions.form.change('buySellTabStatus', 'status', 'order_history'))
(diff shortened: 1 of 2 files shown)


sixtedemaupeou:
I'm working on the same problem: initializing the checkout form everytime we go from any screen to the checkout. However this has to be done in many places and adding that action in all these places doesn't seem like a good solution as we could easily forget one.
I see other solutions that should fix all related issues:

1. Registering a saga on the `COINIFY_NEXT_CHECKOUT_STEP` action type that checks is the step is checkout and initialises the form. The downside is that we would need to pass the form type (buy or sell) to the `coinifyNextCheckoutStep` action
2. Making OrderCheckout an actual React Component instead of a function and initialise the form in the componentDidMount. The downside is the amount of refactoring involved.
3. Use componentWillReceiveProps (soon deprecated so probably not the best solution) in both Buy and Sell index to check if the step prop changes and if nextProps.step is checkout and initialize the form. But that means duplicate code and using a soon deprecated lifecycle method.
What do you think is the best way to do this?

prwelber:
Good points all around. I'm wondering though - where else do we need initialization?

Currently, it happens when buy and sell mount which covers the vast majority of cases. When the signup modal closes is covered by this PR. Where else do we need it? 

When moving back to that step via the stepper, it doesn't need initialization. And when you switch tabs, the component remounts, triggering the initialization.

Thanks for giving this some extra thought and coming up with good solutions! 🙌

sixtedemaupeou:
We need initialization everytime we get to Buy or Sell and so Buy or Sell mounts and then, staying inside Buy or Sell we leave the checkout before going back to it. So everytime we start the buy or sell flow and go back/cancel before completing it.
Good point about `QuoteInput`, I hadn't noticed that it was a React component. It seems like the right solution

PR #553, opened 2018-06-13
respect invited field in settings

salomegeo:
## Description
954

## Change Type
- Bug Fix

## Testing Steps
Use v2 wallet to log in.

## Code Checklist


salomegeo:
I'm not sure this is the way to go, do we even need this invited field? is it written somewhere else in v2-settings? if so upgrade could read it from there and put it in the settings.

plondon:
We need it to roll out partnerships. It should be in get-info, not sure why not if its after you upgrade though.

plondon:
Same issue in v3

plondon:
Just added a note to self.

PR #568, opened 2018-06-14
fix(Sign Message): sign msg from uncompressed key

plondon:
## Description
802

## Change Type
- Bug Fix

## Testing Steps
Sign msg from 5KbC1dvYq3LtDe2xyFPnMcDswxCkAVzAJ1SioRJ3AfBfxNPh3d2 key and 1Ma2DrB78K7jmAwaomqZNRMCvgQrNjE2QC addr

## Code Checklist

### DIFF
--- a/packages/blockchain-wallet-v4/src/signer/btc.js
+++ b/packages/blockchain-wallet-v4/src/signer/btc.js
@@ -44,8 +44,9 @@ export const signWithWIF = curry((network, selection) =>
   compose(signSelection(network), wifToKeys(network))(selection)
 )
 
-export const signMessage = (priv, message) => {
+export const signMessage = (priv, addr, message) => {
   const keyPair = privateKeyStringToKey(priv, 'base58')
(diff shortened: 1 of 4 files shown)


packages/blockchain-wallet-v4/src/signer/btc.js
```diff
@@ -44,8 +44,9 @@ export const signWithWIF = curry((network, selection) =>
   compose(signSelection(network), wifToKeys(network))(selection)
 )
 
-export const signMessage = (priv, message) => {
+export const signMessage = (priv, addr, message) => {
   const keyPair = privateKeyStringToKey(priv, 'base58')
```
salomegeo: Can you move this to privateKeyStringToKey method, because that's used by show private key too, and I think we should be displaying the uncompressed key there too.

plondon: Good point, will do

PR #577, opened 2018-06-14
Label reserved addresses for buy/sell

prwelber:
## Description

## Change Type
- Feature

## Testing Steps
Create Buy trade --> go to wallet addresses --> see label with trade ID

## Code Checklist


packages/blockchain-wallet-v4-frontend/src/data/modules/sfox/sagas.js
```diff
@@ -121,6 +122,22 @@ export default ({ coreSagas }) => {
     }
   }
 
+  const prepareAddress = function * () {
+    try {
+      const state = yield select()
+      const defaultIdx = yield select(selectors.core.wallet.getDefaultAccountIndex)
+      const receiveR = selectors.core.common.btc.getNextAvailableReceiveAddress(settings.NETWORK_BITCOIN, defaultIdx, state)
```
salomegeo: you're mixing the styles here. you either do yield select(...) for all three or you can do getDefaultAccountIndex(state) on line 128 just like 129 and 130. I prefer the former.

prwelber: Good call 👍

PR #582, opened 2018-06-15
Feature/misc bugs6

Lyncee59:
## Description
- Various bug fixes
- Implementation notifications settings

## Change Type
- Bug Fix
- Feature

## Testing Steps
- Test if notifications can be toggled, persist after login again between v3/v4, and that you receive those notifications by sms/email.

## Code Checklist


packages/blockchain-wallet-v4-frontend/src/assets/locales/en.json
```diff
@@ -631,7 +631,8 @@
   "modals.sencbtc.maximumamountmessage": "Not enough funds. Use",
   "modals.sencbtc.maximumfeemessage": "Unnecessarily high fee. Use",
   "modals.sencbtc.minimumamountmessage": "Below minimum amount. Use",
-  "modals.sencbtc.minimumfeemessage": "Low custom fee not recommended. Use",
+  "modals.sencbtc.minimumfeemessage": "Low fee not recommended. Use",
+  "modals.sencbtc.minimumonesatoshimessage": "Invalid (< 1sat/b)",
```
schnogz: is "(< 1sat/b)" translatable? or should that just be in its own Text component, untranslated.?

packages/blockchain-wallet-v4-frontend/src/modals/SendBitcoin/FirstStep/template.success.js
```diff
@@ -27,12 +27,8 @@ const ColLeft = styled.div`
   align-items: flex-start;
   width: 50%;
 `
-const ColRight = ColLeft.extend`
-  display: flex;
-  flex-direction: column;
-  justify-content: center;
+const ColRight = styled(ColLeft)`
```
sixtedemaupeou: The naming is probably not ideal here

packages/blockchain-wallet-v4/src/redux/settings/sagas.js
```diff
@@ -193,6 +185,26 @@ export default ({ api }) => {
...
+    const sharedKey = yield select(wS.getSharedKey)
+    const response = yield call(api.enableNotifications, guid, sharedKey, value)
+    if (!contains('updated', response)) { throw new Error(response) }
+    yield put(actions.setNotificationsOn(value))
+  }
+
+  const setNotificationsType = function * ({ types }) {
+    let type = 0
+    if (prop('email', types)) { type = type + 1 }
+    if (prop('mobile', types)) { type = type + 32 }
```
sixtedemaupeou: What do these magic numbers 1 and 32 mean?

packages/blockchain-wallet-v4-frontend/src/modals/TransactionReport/services.js
```diff
@@ -1,29 +1,33 @@
-import moment from 'moment'
+import moment from 'services/MomentHelper'
 
 export const isDatePosteriorToBtcGenesis = selectedDate => {
```
sixtedemaupeou: All the functions isSomething(Btc/Bch)Something(...) should be more generic (eg: isSomethingSomething(..., coin) ) because we don't want to add a new function everytime we add a coin.

packages/blockchain-wallet-v4/src/redux/settings/sagas.js
sixtedemaupeou: Shouldn't it be 2 instead of 32 btw?

packages/blockchain-wallet-v4-frontend/src/assets/locales/en.json
Lyncee59: mmm yep you are right, we can't translate that, so it should go in as a dynamic value

packages/blockchain-wallet-v4-frontend/src/modals/TransactionReport/services.js
Lyncee59: Mmm I understand the point, but if the function was generic, it would go in the FormHelper instead of staying here, we'll have to factorize more the validation rules once we reached some consensus/conventions on how to pass the props to our form.

Lyncee59: For the comment below regarding the mobile value, no its basically in bytes: 10000 which is 32

packages/blockchain-wallet-v4/src/redux/settings/sagas.js
Lyncee59: Nope, basically its a byte system: 1byte 91) is email, 10000bytes (32) for mobile, so you can use 10001 (33) if you want to activate both

packages/blockchain-wallet-v4-frontend/src/modals/SendBitcoin/FirstStep/template.success.js
Lyncee59: Yea well it has been there for ages, agree the naming is not great ! it should only be called Column maybe and pass a specific prop to change the alignment. Will do another time !

packages/blockchain-wallet-v4-frontend/src/assets/locales/en.json
Lyncee59: updated

PR #600, opened 2018-06-18
feat(coinify: poll for kyc update from ISX)

sixtedemaupeou:
## Description

## Change Type
- Feature
- Bug Fix
- CI/CD
- Upgrades
- Other (add explanation)

## Testing Steps

## Code Checklist


packages/blockchain-wallet-v4/src/redux/data/coinify/sagas.js
```diff
@@ -303,6 +304,27 @@ export default ({ api, options }) => {
...
+        const sortedKYCs = sort(byTime, kycs)
+        status = statusPath(sortedKYCs)
+        yield put(A.getKYCsSuccess(sortedKYCs))
+      }
+    } catch (e) {
+      console.log('pollKYCsPending failure', e)
+      yield put(A.getKYCsFailure(e))
+    }
+  }
+
```
plondon: Can we extract this out to a more generic polling function?

```
function poll(check, success) {
   if (!check) keepPolling()
   else success()
}```

Comment (sixtedemaupeou) [reply packages/blockchain-wallet-v4/src/redux/data/coinify/sagas.js]: We could do that but I'm not sure it would be very beneficial given:
-a while loop wouldn't be much more complex than that function
-sometimes we might want to run the keepPolling function at least once (ie do while) and sometimes not
-it means we need to create another (generator) function for keepPolling (here we don't need any success function)
Also, we would need to pass the delay as an argument because we'll want different delays for different usecases of the poll function (if we ever need to use that function in other places

Comment (plondon) [reply packages/blockchain-wallet-v4/src/redux/data/coinify/sagas.js]: Okay, we can revisit this in the future if we need multiple polling functions and they are all similar.

**Comment by prwelber** [comment]
Here's the call we should use for this so we're not refetching all KYC's when polling

https://github.com/blockchain/bitcoin-coinify-client/blob/master/src/kyc.js#L65

File: packages/blockchain-wallet-v4/src/redux/data/coinify/sagas.js [hunk 2]
```diff
@@ -295,47 +295,51 @@ export default ({ api, options }) => {
...
         yield call(delay, 1000)
-        const sortedKYCs = yield call(sortKycs)
-        status = statusPath(sortedKYCs)
-        yield put(A.getKYCsSuccess(sortedKYCs))
+        const kycR = yield select(S.getKyc)
+        const kyc = kycR.getOrElse(undefined)
+        if (!kyc) {
+          return
+        }
+        yield apply(kyc, kyc.refresh)
```
prwelber: 🎉

prwelber:
@sixtedemaupeou Looks good! Can you please fix the build before merging? It's just a couple of unused imports.

packages/blockchain-wallet-v4/src/redux/data/coinify/sagas.js
```diff
@@ -295,47 +295,51 @@ export default ({ api, options }) => {
...
-  const getKYCs = function * () {
+  const getKYC = function * () {
     try {
-      yield put(A.getKYCsLoading())
-      const sortedKYCs = yield call(sortKycs)
-      yield put(A.getKYCsSuccess(sortedKYCs))
-      return sortedKYCs
+      yield put(A.getKYCLoading())
+      const coinify = yield call(getCoinify)
+      const kyc = yield call(getLastKyc, coinify)
```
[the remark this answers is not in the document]
sixtedemaupeou: Good point. `getLastKyc` used to be a function used by both `getKYC` and `pollKYCPending` but I removed it from the latter so it does not need to be its own function anymore. Just moved it back inside `getKYC`

PR #603, opened 2018-06-18
fix(backup phrase validation)

schnogz:
## Description
- Fixes backup phrase validation as per ticket 1029. 
- Fixes form errors placement as per ticket 517.
- Cleans up and refactors various parts of security center code.

## Change Type
Bug Fix

## Testing Steps
See bug descriptions.

## Code Checklist


salomegeo:
validator still doesn't work though :(, like if the text field is empty required should be displayed, but it's not. If qa says this is ok, i'll approve it. sigh

schnogz:
It's not going to display the red "required" warning until you give focus to the input and then unfocus without entering a value. Why would we yell at them right away?

PR #618, opened 2018-06-19
bch should accept btc address

salomegeo:
## Description
973

## Change Type
- Bug Fix

## Code Checklist

### DIFF
--- a/packages/blockchain-wallet-v4-frontend/src/components/QRCodeCapture/index.js
+++ b/packages/blockchain-wallet-v4-frontend/src/components/QRCodeCapture/index.js
@@ -49,7 +49,7 @@ class QRCodeCaptureContainer extends React.PureComponent {
 
   handleScanBchAddress (data) {
     try {
-      if (utils.bch.isCashAddr(data)) {
+      if (utils.bitcoin.isValidBitcoinAddress(data)) {
         this.props.formActions.change('sendBch', 'to', data)
         this.props.updateUI({ bchAddress: { toggled: false } })
         return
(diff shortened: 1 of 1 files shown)


packages/blockchain-wallet-v4-frontend/src/components/QRCodeCapture/index.js
```diff
@@ -49,7 +49,7 @@ class QRCodeCaptureContainer extends React.PureComponent {
 
   handleScanBchAddress (data) {
     try {
-      if (utils.bch.isCashAddr(data)) {
+      if (utils.bitcoin.isValidBitcoinAddress(data)) {
```
schnogz: You are calling a BTC addr validation method when the outer function hah BCH in it's name. Is this intended?

salomegeo: yes. bch should accept btc address.

schnogz: So then it should be, `utils.bitcoin.isValidBitcoinAddress(data) || utils.bch.isCashAddr(data)`? Or no.

salomegeo: no because isCashAddr("bitcoincash:qpheqxjud2p0yg85t679tnc7n05xffs5ks07zhg6y6") evaluates to true, take a look at the screenshot in the issue.

so for bitcoin isValidBitcoinAddress returns true for plain addresses only and decode only works for prefixed address strings. for bitcoin cash both decode and isCash works for both, but decode works better because it correctly parses the address.

for bitcoin cash, we need to handle the case where the address is bitcoin address.

plondon: So later on down in the function is where bch is handled correctly. Sorry for the confusion

PR #622, opened 2018-06-20
pass second password for eth metadata creation

salomegeo:
## Description
1043

## Change Type
- Bug Fix

## Code Checklist


schnogz:
Can we please add a comment in the code explaining the need for this 2nd ask and our need for future refactor? Without it, someone will delete the code and bug will be back.

salomegeo:
this will only prompt for second password when we're creating kvstore entry for the first time and second password is already enabled. this can only happen when we're doing v2-v4 upgrade and it was already asking for the second password for like 5 times so it doesn't make that much difference. It would be ideal if that number was 1.

salomegeo:
Added a comment @schnogz.

Honestly I don't think it would be that difficult to cut down on number of second password propms for v2-v4 upgrade, but if you haven't logged in since v2 you probably never will.

PR #641, opened 2018-06-22
Qr error message

salomegeo:
## Description
Display private key doesn't match on the bottom, and also force validate by calling touch. right now if you enter a valid bitcoin private key  that doesn't match the watch only public key, the error isn't displayed unless you click on the text box, that causes some other problems.

isCashAddr is necessary after all, you were right. but it has to run after ```bip21.decode(data, 'bitcoincash')```

## Change Type
- Bug Fix
- CI/CD

## Code Checklist

### DIFF
--- a/packages/blockchain-wallet-v4-frontend/src/components/QRCodeCapture/index.js
+++ b/packages/blockchain-wallet-v4-frontend/src/components/QRCodeCapture/index.js
@@ -49,7 +49,7 @@ class QRCodeCaptureContainer extends React.PureComponent {
 
   handleScanBchAddress (data) {
     try {
-      if (utils.bitcoin.isValidBitcoinAddress(data)) {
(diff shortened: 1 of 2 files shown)


packages/blockchain-wallet-v4-frontend/src/components/QRCodeCapture/index.js
```diff
@@ -62,6 +62,11 @@ class QRCodeCaptureContainer extends React.PureComponent {
       this.props.updateUI({ bchAddress: { toggled: false } })
     } catch (e) {
       try {
+        if (utils.bch.isCashAddr(data)) {
+          this.props.formActions.change('sendBch', 'to', data)
+          this.props.updateUI({bchAddress: {toggled: false}})
+          return
+        }
```
plondon: Could this be simplified with a `utils.bch.isCashAddr(data) || utils.bitcoin.isValidBitcoinAddress(data)` in the first try? Think Andrew mentioned it last time also

PR #650, opened 2018-06-22
Feature/translation mgmt

plondon:
- [x] Todo: figure out what to do with `SanitizedFormattedHTMLMessage`

## Description
Translation mgmt

## Change Type
- Upgrades

## Testing Steps

## Code Checklist


plondon:
CC @kristovatlas too.

I've removed SanitizedFormattedHTMLMessage as it wasn't actually doing anything. react-intl doesn't let us use dynamic strings, as Guillaume has pointed out before, so not sure why I thought this would work, that translations were never picked up.

FormattedHTMLMessage does use dangerouslySetInnerHTML but strips `<script>` tags and our csp blocks inline execution, so it may be safe, but will defer to Kristov

packages/blockchain-wallet-v4-frontend/translationRunner.js
```diff
@@ -0,0 +1,52 @@
...
+  messagesDirectory: './build/extractedMessages',
+  translationsDirectory: './src/assets/locales',
+  whitelistsDirectory: './src/assets/locales/whitelists',
+  singleMessagesFile: true,
+  languages: [
+    'en'
+  ]
+})
+
+bcManageTranslations({
```
plondon: I modified manageTranslations slightly so instead of outputting the english default message it outputs `null` if the string is empty.

PR #653, opened 2018-06-23
Feature/responsive

prwelber:
## Description
724

## Change Type
- Feature

## Testing Steps
Go through the buy/sell flow using dev tools device toolbar

## Code Checklist


prwelber:
Done making commits to this branch. Can be reviewed/merged.

packages/blockchain-wallet-v4-frontend/src/services/ResponsiveService/index.js
```diff
@@ -0,0 +1,40 @@
...
+  desktop: 2560
+}
+
+// Iterate through the sizes and create a media template
+const media = Object.keys(sizes).reduce((acc, label) => {
+  acc[label] = (...args) => css`
+    @media (max-width: ${sizes[label]}px) {
+      ${css(...args)}
+    }
+  `
```
plondon: Leora tried to use `min-width` in most places. Using `max-width` is fine but we can't have a combo of min-width and max-width at the same breakpoints. If you're going to use `max-width` decrease the sizes by one px, otherwise some things will break at 768 thinking its mobile, while min-width will keep it unbroken.

Also leora set 480 for mobile breakpoint, let's keep that.

max-width breakpoints:
424,
767,
1023

Do we need laptopL and desktop? They are not used anywhere yet.

prwelber: Breakpoints have been updated. I think all media queries should be cleaned up to use this service at some point, but that's not a priority at the moment. I kept laptopL and desktop for now as making the wallet totally responsive beyond mobile phones is a bit of a WIP.

PR #654, opened 2018-06-23
chore(BuySell): swap receiveAddress

prwelber:
## Description
@plondon what do you think is the best way to get this commit onto the production version yet not break staging next time we do a release? (without manually switching it back and forth)

## Change Type
- Feature
- Bug Fix
- CI/CD
- Upgrades
- Other (add explanation)

## Testing Steps

## Code Checklist

### DIFF
--- a/packages/blockchain-wallet-v4/src/exchange/delegate/index.js
+++ b/packages/blockchain-wallet-v4/src/exchange/delegate/index.js
@@ -1,4 +1,5 @@
 import { btc } from '../../redux/common/selectors'
+import { getDefaultAccountIndex } from '../../redux/wallet/selectors'
 import Bitcoin from 'bitcoinjs-lib'
 
 export class ExchangeDelegate {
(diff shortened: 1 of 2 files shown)


packages/blockchain-wallet-v4/src/exchange/delegate/index.js
```diff
@@ -86,10 +86,9 @@ export class ExchangeDelegate {
   }
 
   reserveReceiveAddress () {
-    // const receiveAddress = btc.getNextAvailableReceiveAddress(Bitcoin.networks.bitcoin.NETWORK_BITCOIN, 0, this.state)
+    const receiveAddress = btc.getNextAvailableReceiveAddress(Bitcoin.networks.bitcoin.NETWORK_BITCOIN, 0, this.state)
```
plondon: Sorry for just noticing this, but this'll always pick your first account, even if default account is index > 0

plondon: I think we can import selectors `core.wallet.getDefaultAccountIndex(this.state)`

plondon: Same issue here actually: https://github.com/blockchain/blockchain-wallet-v4-frontend/blob/a6a1bd83de77c43370a8348d7cb82b121052a594/packages/blockchain-wallet-v4-frontend/src/scenes/BuySell/SfoxCheckout/Content/selectors.js#L9

plondon:
> @plondon what do you think is the best way to get this commit onto the production version yet not break staging next time we do a release? (without manually switching it back and forth)

I will just tag a new release for alpha15, release that build to production, but not release to staging.

packages/blockchain-wallet-v4/src/exchange/delegate/index.js
prwelber: Good call. Fixed.

PR #687, opened 2018-06-26
hotfix(exchange delegate: check environment for address generation)

sixtedemaupeou:
## Description

## Change Type
- Bug Fix

## Testing Steps

## Code Checklist


sixtedemaupeou:
The logic is not perfect here but probably good enough. I had to assume that if any of coinify or sfox production flags in wallet options is set to false we want to use a testnet address for both

PR #690, opened 2018-06-26
Handle SFOX profile failure

prwelber:
## Description
696

## Change Type
- Bug Fix

## Testing Steps
Fail the address/id verification step and close the tray. Failure view should not render.

## Code Checklist


prwelber:
We call verify on the profile and if sfox returns an error, we set the profile to a failed state. https://github.com/blockchain/blockchain-wallet-v4-frontend/blob/development/packages/blockchain-wallet-v4/src/redux/data/sfox/sagas.js#L184

One alternative it to create a `profileVerified` entry in the state or something separate from the main `profile` entry, but I don't really see that as a better solution. It becomes another piece of state to keep track of during the signup flow.

PR #692, opened 2018-06-27
Fix/notifications wallet sync

tony-blockchain:
## Description
removed update-notifications-on call, added active to sync request payload

## Change Type
- Bug Fix

## Testing Steps
1. Activate sms in preferences on target wallet
2. Send funds to target wallet
3. Receive sms notification that funds arrived on the wallet's phone number

## Code Checklist


plondon:
We should do a write up of this for a wiki article so our findings don't get lost

packages/blockchain-wallet-v4/src/redux/walletSync/middleware.js
```diff
@@ -1,18 +1,88 @@
...
+ */
+export const unusedAddressesCount = 10
+
+/**
+ * Launches derivation of future addresses for target HDAccount
+ * @curried
+ * @param {any} state redux state
+ * @param {HDAccount.HDAccount} account
+ * @returns {Promise<String>[]} array of promises, each of them resolves into address
+ */
```
pernas: I never get used to that kind of annotation. I think they are super noisy. 
I tend to use the type annotation:
// getHDAccountAddressPromises :: state -> Account -> [Promise(Address)]

packages/blockchain-wallet-v4/src/redux/walletSync/middleware.js
```diff
@@ -1,18 +1,88 @@
...
+          HDAccount.getReceiveAddress(
+            account,
+            index,
+            networks.bitcoin.NETWORK_BITCOIN
+          )
+        ), 0)
+      ))
+    ))
+    .getOrElse([])
+})
```
pernas: Several comments here:

I would avoid writing comments explaining your code. I prefer simpler code and less comments.
For instance, the line using converge is infamous, you are just creating a range from A to A+B. So it is equivalent to `range(A, A + B)`. Actually, in the comment itself you have the equivalent simple code. 
Also, I would avoid defining long functions (several lines), like the one doing the setTimeout, in the middle of your composition chain.

I would also avoid using pipes and maps together. In the end, I would have wrote something like:

```
const toAsync = (f) => new Promise(
    (resolve) => setTimeout(() => resolve(f()), 0)
  )

const getHDAccountAddress = curry((state, account) => {
  const xpub = selectors.wallet.getAccountXpub(account.index, state)
  return selectors.data.bitcoin.getReceiveIndex(xpub, state)
    .map(ri => range(ri, ri + unusedAddressesCount))
    .getOrElse([])
    .map(i => toAsync(HDAccount.getReceiveAddress.bind(null, account, index, networks.bitcoin.NETWORK_BITCOIN)))
})
```  
The journey it is a lot clearer to me here:

Remote.of(23) --> Remote.of([23, 24, ..., 32]) . --> [23, 24, ..., 32] --> [Promise("1addr1"), Promise("1addr2", ...)]

and you can see these 4 steps in the four lines clearly separated.

tony-blockchain: Some bad code form me, thanks for improving readability.

The setTimeout comment is required here as it descries concepts that are not obvious to anyone except maybe v8/browser developers.
I'd personally try to replace setTimeout with requestAnimationFrame on batch actions, because of superior performance. And in this case such replacement is bad which is not obvious.

And one letter variables are not an option IMO, even in small functions, because they are a big pain to search for.

pernas: About the one letter parameters. I guess it is a matter of opinion. I love them. Those things are defined locally to only be used on the same line and they don't have any meaning in any outer scope, so I cannot see how you would need to search for the variable inside that block. If you need to search from something you would look for the outer function which is what is meaningful in the outer context. On the other side, short names improve a lot the readability. SpeciallyBecauseReadingSentencesWrittenThatWayIsNotThatEasy.

tony-blockchain: depends on context I guess, f is ok, however receiveIndex is quite searchable

pernas: Indeed, this is why the function ReceiveIndex is called that way and not ri. This is the important element to care about from an outer perspective further away of those 3 lines of code. And if you search for that will find that piece of code.

PR #695, opened 2018-06-27
Top Balance: hide separator when no watch only balances; other fixes

prwelber:
Because of the way these components are organized, there wasn't a straight forward way to do this since you need to have all watch only balance data in the same place to determine showing/hiding the separator. The other option probably means reshuffling components and selectors around, and I'd rather not at this point. Open to suggestions if there's a better way.

## Description
Hide the `separator` component when there's no btc and bch watch only balances.

## Change Type
- Bug Fix

## Testing Steps

## Code Checklist

### DIFF
--- a/packages/blockchain-wallet-v4-frontend/src/layouts/Wallet/MenuTop/Balance/BchWatchOnlyBalance/index.js
+++ b/packages/blockchain-wallet-v4-frontend/src/layouts/Wallet/MenuTop/Balance/BchWatchOnlyBalance/index.js
@@ -18,6 +18,10 @@ class BchWatchOnlyBalance extends React.PureComponent {
     this.props.actions.fetchData()
   }
 
+  componentDidUpdate () {
+    this.props.data.getOrElse() && this.props.hasBchSubBalance()
+  }
+
   handleRefresh () {
     this.props.actions.fetchData()
   }
--- a/packages/blockchain-wallet-v4-frontend/src/layouts/Wallet/MenuTop/Balance/BtcWatchOnlyBalance/index.js
+++ b/packages/blockchain-wallet-v4-frontend/src/layouts/Wallet/MenuTop/Balance/BtcWatchOnlyBalance/index.js
@@ -18,6 +18,10 @@ class BtcWatchOnlyBalance extends React.PureComponent {
     this.props.actions.fetchData()
   }
 
+  componentDidUpdate () {
+    this.props.data.getOrElse() && this.props.hasBtcSubBalance()
+  }
+
   handleRefresh () {
     this.props.actions.fetchData()
   }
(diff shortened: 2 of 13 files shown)


packages/blockchain-wallet-v4-frontend/src/layouts/Wallet/MenuTop/Balance/BchWatchOnlyBalance/index.js
```diff
@@ -18,6 +18,10 @@ class BchWatchOnlyBalance extends React.PureComponent {
     this.props.actions.fetchData()
   }
 
+  componentDidUpdate () {
+    this.props.data.getOrElse() && this.props.hasBchSubBalance()
+  }
```
plondon: ?

packages/blockchain-wallet-v4-frontend/src/services/ResponsiveService/index.js
```diff
@@ -3,7 +3,7 @@ import { css } from 'styled-components'
 const sizes = {
   mobile: 479,
   tablet: 767,
-  laptop: 1023,
+  laptop: 1200,
```
plondon: 1199

packages/blockchain-wallet-v4-frontend/src/layouts/Wallet/MenuTop/Balance/BtcWatchOnlyBalance/index.js
```diff
@@ -18,6 +18,10 @@ class BtcWatchOnlyBalance extends React.PureComponent {
     this.props.actions.fetchData()
   }
 
+  componentDidUpdate () {
+    this.props.data.getOrElse() && this.props.hasBtcSubBalance()
+  }
```
plondon: This seems a bit backwards. It tells the parent component about a boolean, when the parent component itself could know

plondon: I guess fine

prwelber: The parent can't know about it because of where the selector is located. I'll move things around to make this more clear.

prwelber:
@plondon lots of non-trivial changes in this commit, where we get the watch only balances at the top of the component tree and pass them down into the template.  This interferes with error handling for individual currencies, which I don't like. https://github.com/blockchain/blockchain-wallet-v4-frontend/pull/695/commits/d7a1af9de2a8a165d901b808d1e3f239279befd2

After doing it this way, I prefer the first approach even though it has it's own flaws.

PR #702, opened 2018-06-27
fix(Eth): transferred/sent tx fees

plondon:
## Description
1060

## Change Type
- Bug Fix

## Testing Steps
All currencies sent txs: should show amount sent + fee
All currencies received txs: should show amount received (NO fee)
**BCH/BTC/ETH transferred txs: should show amount sent (NO fee)**

## Code Checklist

### DIFF
--- a/packages/blockchain-wallet-v4/src/transactions/ethereum.js
+++ b/packages/blockchain-wallet-v4/src/transactions/ethereum.js
@@ -39,17 +39,19 @@ export const getLabel = (address, state) => {
 
 export const _transformTx = curry((addresses, latestBlock, state, tx) => {
   const fee = getFee(tx)
+  const type = toLower(getType(tx, addresses))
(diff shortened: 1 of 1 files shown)


plondon:
Adds fee to total value for only sent txs

packages/blockchain-wallet-v4/src/transactions/ethereum.js
```diff
@@ -39,17 +39,19 @@ export const getLabel = (address, state) => {
 
 export const _transformTx = curry((addresses, latestBlock, state, tx) => {
   const fee = getFee(tx)
+  const type = toLower(getType(tx, addresses))
```
sixtedemaupeou: Why use toLower here when you could simply return 'transferred', 'sent', 'received' or 'unknown' all lowercase in the getType function above

PR #703, opened 2018-06-28
style(Top Balance): dont show separator if no unspendable balances

plondon:
Small style change to not show unspendable balance separator if balances are 0


packages/blockchain-wallet-v4-frontend/src/layouts/Wallet/MenuTop/Balance/BchWatchOnlyBalance/template.success.js
```diff
@@ -22,7 +22,7 @@ const Wrapper = styled.div`
 `
 
 const Success = props => {
-  return props.balance === 0 ? <div /> : (
+  return props.balance === 0 ? null : (
```
sixtedemaupeou: I believe you could even have props.balance === 0 && (<LinkContainer...>...</LinkContainer>)

PR #709, opened 2018-06-28
fix(kvstore buysell selectors remote)

sixtedemaupeou:
## Description
Fix buy sell metadata selector: Do not take data of Remote but expose Remote

## Change Type
- Bug Fix

## Testing Steps

## Code Checklist


sixtedemaupeou:
Considering the buysell metadata is a Remote, its data may or may not be there so in the selector we don't want to use the data field to access what's in it as it would give us something that's not a Remote, implying that the data we're getting is there when actually it's not there.
If you look at these two lines, if the buysell metadata was Remote.Loading or Remote.Failure(e) kvTrades would have been undefined and on the next line calling kvTrades.length would have thrown an error.
```const kvTrades = yield select(buySellSelectors.getSfoxTrades)
const numberOfTrades = kvTrades.length```

PR #711, opened 2018-06-28
Sfox fix

prwelber:
## Description
696

## Change Type
- Bug Fix

## Testing Steps
See ticket

## Code Checklist


packages/blockchain-wallet-v4-frontend/src/modals/SfoxExchangeData/Link/index.js
```diff
@@ -44,6 +44,11 @@ class LinkContainer extends Component {
     }
   }
 
+  componentWillUnmount () {
```
prwelber: reset the component UI on unmount and wipe the bank accounts from the redux state. We do this because we lose the token that's needed for linking a bank when the component's state is destroyed.

packages/blockchain-wallet-v4/src/redux/data/sfox/reducers.js
```diff
@@ -87,6 +87,9 @@ const sfoxReducer = (state = INITIAL_STATE, action) => {
     case AT.GET_BANK_ACCOUNTS_FAILURE: {
       return assoc('bankAccounts', Remote.Failure(payload), state)
     }
+    case AT.WIPE_BANK_ACCOUNTS: {
+      return assoc('bankAccounts', null, state)
```
sixtedemaupeou: You should use Remote.NotAsked instead of null

prwelber: Good call, fixed in latest commit

PR #725, opened 2018-07-02
fix - refresh fixed missing BTC/BCH data in request modal

tony-blockchain:
## Description
Refetching data in BTC/BCH request forms if one is missing.

## Change Type
- Bug Fix

## Testing Steps
1. Grab a state from 1108 issue
2. Load it while logged in or twice
3. Open request dialog - get an error ( loading state while logged out yields no error message)
4. Click on refresh
5. Should show a request window

## Code Checklist


tony-blockchain:
eth is not an issue as intialData is just coin: 'ETH', let me look at the logic

tony-blockchain:
and reinit looks fine as it's needed if redux-form state goes trashed.
fetchData triggers reinit itself through componentDidUpdtae

PR #726, opened 2018-07-02
fix - refetching all transactions for activity stream while on the home page

tony-blockchain:
## Description
Home page activity stream is not refreshed without transaction refetch

## Change Type
- Bug Fix

## Testing Steps
1. Make a transaction that won't show up in the activity stream, btc state from 1108 is valid
2. Click refresh while on the main page
3. See activity stream refreshed

## Code Checklist


plondon:
Was there a ticket number for this? Just curious so I can add it to the sprint if yes

tony-blockchain:
not yet, did not make an issue afair

plondon:
Ok no need if the work is already done. Thanks

Issue #710, opened 2018-06-28
3 child repos do not have package locks


kristovatlas:
# 🐛 Bug Report 🐛

## Description
No `package-lock.json` file is present in the following 3 sub-repos:
* https://github.com/blockchain/blockchain-wallet-v4-frontend/tree/development/packages/blockchain-info-components
* https://github.com/blockchain/blockchain-wallet-v4-frontend/tree/development/packages/blockchain-wallet-v4-frontend
* https://github.com/blockchain/blockchain-wallet-v4-frontend/tree/development/packages/blockchain-wallet-v4

Our build process starts with `yarn bootstrap` from the parent repo directory. The parent repo directory has its own `package-lock.json` file, but it doesn't appear this will be respected in the 3 sub-repo's node modules during building.

## Reproduction Steps
This is outside of the normal building process, but if you believe that lerna is using npm to build the sub-repos, then you can recreate this using `npm install` from the `blockchain-info-components directory`. `yarn` doesn't provide any warnings.

## Additional Information
* https://stackoverflow.com/questions/44206782/do-i-commit-the-package-lock-json-file-created-by-npm-5
* https://docs.npmjs.com/files/package-lock.json

kristovatlas:
Same issue for the other 2 sub-repos.

After talking with @schnogz, there's no obvious quick fix.

Although we use `yarn` to build, we're not using its locking system. `yarn bootstrap` calls lerna to run the installs, using npm in turn. Therefore, the top-level package-lock.json file will be respected, but it doesn't appear that it will be applied to the 3 sub-repos with their own `package.json` files.

Resolution: There may be a way in lerna to generate the child lock files. The dev team is also considering getting rid of lerna and using npm's workspaces functionality, which should simply the package locking situation.

tony-blockchain:
1. We have to switch to yarn workspaces, which improves build time, by saving time downloading common modules
2. I'd migrate to yarn.lock and using yarn as installer, as in my experience it proved to be much more stable and well designed. Npm has changed their lock behaviour twice already:
- First time they've built from package-lock
- Then(npm v5) they started updating package-lock during install (wth?)
- Now they introduced `npm ci` which builds from package-lock

tony-blockchain:
Here's the original workspaces issue related to child yarn.locks
https://github.com/yarnpkg/yarn/issues/5428

[...]

PR #736, opened 2018-07-05
upgrade - npm => yarn, lerna bootstrap => yarn workspaces

tony-blockchain:
## Description
Yarn is faster at building than npm/lerna

However there's no clean way yet to run command in every package using yarn workspaces.
There's https://github.com/whoeverest/wsrun but I wouldn't trust such that such package cannot introduce a vulnerability compared to yarn/lerna (lerna also shouldn't be blindly trusted ofc)
So for now I'd keep lerna for the sole purpose of running task among every package.

To install packages run `yarn` in root. To add module do a `cd packages/package-name` and `yarn add module-name` 

## Change Type
- Upgrades

## Testing Steps
Test commands you usually use.
I've tested build/start/test

## Code Checklist


plondon:
> So for now I'd keep lerna for the sole purpose of running task among every package.

But this PR removes it

tony-blockchain:
lerna remains to launch jobs in every package, I now think that it's better to replace it by manual calling each package

schnogz:
If we wait until the components package is busted out to its own repo, we shouldn't need Lerna to run commands. Just manually calling the two remaining packages wont be too bad. Let's leave this PR open until that happens.

tony-blockchain:
@schnogz it isn't any worse just calling 3 packages manually right. Amount of packages won't increase anyway and the repo split's not a quick thing, so I guess I wipe lerna config and ci install script, change a couple of scripts and we're good to go.

tony-blockchain:
checked `wsrun` with `snyk` and looked over the code myself, looks safe.
from now on it's fully yarn, nothing else involved

schnogz:
This PR will address issue #710 !

PR #755, opened 2018-07-09
Updates to NGINX 

schnogz:
## Description
- Adds support for languages in URL path
- Adds support for iFrame migration magic by adding `rootURL` to `frame-src` in CSP header
- Changes NGINX to not return 404 on unmatched route and instead let React Router handle it

## Change Type
Feature, Bug Fix

## Testing Steps
Application routing works in development/staging environments

## Code Checklist


plondon:
> Adds support for iFrame migration magic by adding rootURL to frame-src in CSP header
Thanks!

PR #774, opened 2018-07-13
Fix/2fa links

plondon:
## Description
1129

## Change Type
- Bug Fix

## Testing Steps
navigate to reset your 2fa

## Code Checklist

### DIFF
--- a/packages/blockchain-wallet-v4-frontend/src/scenes/Reset2FAToken/index.js
+++ b/packages/blockchain-wallet-v4-frontend/src/scenes/Reset2FAToken/index.js
@@ -20,7 +20,7 @@ const Wrapper = styled.div`
 class Reset2FAToken extends React.PureComponent {
   constructor (props) {
     super(props)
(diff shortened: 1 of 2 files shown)


packages/blockchain-wallet-v4-frontend/src/scenes/Reset2FAToken/index.js
```diff
@@ -20,7 +20,7 @@ const Wrapper = styled.div`
 class Reset2FAToken extends React.PureComponent {
   constructor (props) {
     super(props)
-    this.state = { token: decodeURIComponent(props.location.pathname.split('/reset2fa-token/')[1]) }
+    this.state = { token: decodeURIComponent(props.location.pathname.split('/reset-two-factor/')[1]) }
```
plondon: Its a bit confusing, there are 2 reset two factor auth links. One is in the wallet (reset-2fa) and starts the reset process, the second (above, reset-two-factor) is an external link from an email that the wallet parses a token from.

PR #779, opened 2018-07-13
chore(Add Prettier and pre-commit hooks)

schnogz:
## Description
Large PR that conforms all our code to the standards laid out by Prettier.

## Questions
- Thoughts on rules?  I'm lukewarm on some such as double-quotes and semi-colons.
- Pre-push vs pre-commit hooks?  I put all the validation in the pre-push because pre-commit seems too overbearing and time consuming but open to suggestions.
- Should Prettier commands autofix issues? I have it set up to write changes to the files if there are violations. We could set it up to just fail and then the developer would need to run a different command to fix the actual issues.

## Change Type
CI/CD

## Testing Steps
Prettier commands work and pre-push hooks stop origin pushes if there are violations

## Code Checklist


schnogz:
There isnt a prettier rule to force spaces after function declarations. (discussed [here](https://github.com/prettier/prettier/issues/1139)).  So we could do one of the following:
- Install the prettier fork [prettier-misc](https://github.com/arijs/prettier-miscellaneous) and enable
- Enable `space-after-function-name` ESLint rule and run it with `--fix` after prettier has ran. although this will lead us to other problems such as not being able to throw CI build errors on prettier violations
- Dont enforce spaces after function declarations

PR #786, opened 2018-07-16
Release/v4.0.0 beta1

plondon:
Release


plondon:
Ok, this should be correct now

PR #802, opened 2018-07-23
feat(Tooltips): updated tooltips

jstettner:
## Description
Tooltips are now no longer bound by parent divs. Refactored tooltip system allows for more reusable code in the future. 

These new tooltips work by including `<Tooltip id='tooltipId'> {content} </Tooltip>` **higher level**, and then wrapping the label of the tooltip with a TooltipHost component. A hover on these items will now be triggers for any existing <Tooltip /> components where the id prop matches the Tooltip id.

This can allow, for instance, one description Tooltip for coin-send to be used across components for different currencies by mounting it **once** above the send modals, and then have multiple labels with the same id.

The updated Tooltip component takes advantage of the react-tooltip lib which offers more [features](https://github.com/wwayne/react-tooltip/).

TODO:
- Add proptypes and default props to the new Tooltip

## Change Type
- Upgrades

## Testing Steps
Test tooltips everywhere.

## Code Checklist


packages/blockchain-wallet-v4-frontend/src/scenes/Settings/Addresses/Btc/ManageAddresses/UsedAddresses/Table/template.js
```diff
@@ -9,9 +10,14 @@ import {
   TableHeader,
   TableRow,
   TableCell,
-  Tooltip
+  Icon
 } from 'blockchain-info-components'
 
+const TooltipIcon = styled(Icon)`
```
tony-blockchain: TooltipIcon is similar everywhere, should be added as a common component

tony-blockchain:
Ideally I'd also create a HOC/component as tooltip host to have data attributes and call TooltipRebuild() on it's mount and unmount:
- this would reduce the number of calls( no need to call on each update )
- this would encapsulate the call which otherwise we'd need to do manually

jstettner:
@tony-blockchain Ideally the goal would be to use one tooltip for multiple triggers such as the descriptions in all send modals, for instance. For this to work it would have to be placed above the mounting of the modals. In this case it varies. The TxFeedTooltips component is a good example of how this works. Only one tooltip is mounted to the dom and then its position and opacity (and even the content) can change. 

I tried to place the tooltips as high up as I could while still making them component specific. i.e. If the tooltip is triggered by an icon hover on a STEP 1 in a component, it can be mounted above the steps entirely so that it isn't remounted if someone clicks `previous` (step 2 > step 1). The tooltip is mounted the entire time and doesn't need to reload.

tony-blockchain:
if we mount them on the upmost level they won't be in dom but will be available everywhere, right
if we make a separate host HOC for tooltips it can trigger recalculation on both mount and unmount and remove the uneeded tooltips from DOM

jstettner:
@tony-blockchain I'm not sure I fully understand how the HOC would work. We could have a component mounted on the highest level that contains all of the tooltips which are mounted only once. How would it trigger recalculation when an item with data-tip is mounted?

tony-blockchain:
it would trigger `TooltipRebuild` on it's own `componentDidMount` and `componentWillUnmount`

jstettner:
but it would only mount once?

tony-blockchain:
I see the thing, they won't be unmounted, react-tooltip does not provide that

jstettner:
Yes. Ideally a project would contain very few tooltips, if not one, and it would be controlled by attributes on certain items.

jstettner:
The problem we have is our tooltips contain components, not plaintext, which makes it impossible to mount only one. we end up with something similar to the previous tooltip system.

tony-blockchain:
They'll be still better grouped on the upmost level and rebound with `rebuild` automatically upon host mount, that way it would be easier to use them:
1. Add tooltip to the tooltip registry
2. Add host
vs
1. Add tooltip somewhere higher where `componentDidUpdate` is triggered
2. Add a rebuild to the `componentDidUpdate`(actually putting it in cata imo is not  a nice idea, as it would fire on every render)
3. Add the data attributes

tony-blockchain:
Previous system had the problem of tooltip duplication, as far as I understood, this is still going to be solved

jstettner:
The previous system mounted individual tooltips for every location in which they would occur. They were also bound by containers and would often be covered (look at a transaction feed on staging and hover over addresses especially near the top), where these tooltips actually mount outside of the containers and avoid this issue.

jstettner:
Does this work as the final solution: consolidate all tooltips into a component and mount it on the highest level, put the rebuilds in the componentDidUpdate() functions for each component that contains a data-tip item?

tony-blockchain:
Yes, that's almost it! Could you add one more thing:
Instead of `data-tip` attribute, use a HOC which will trigger rebuild upon mount, it will reduce calls to rebuild a lot and we won't need to call it manually

tony-blockchain:
even better not a HOC but just a component

jstettner:
Yeah that's a great idea. So something like a TooltipHost component that would take a component as a prop, would have TooltipRebuild() in it's componentDidMount, and would attach data-tip and data-for attributes?

tony-blockchain:
yes, perfect

PR #803, opened 2018-07-23
[WIP] Add testnet support

lzhuor:
## TODO
- [x] Fix kvstore entries
- [x] Fix send
- [x] Fix add wallet account
- [ ] Disable other currencies?

## Description
Added logic to check Node.js environment in `config/index.js` so Bitcoin Network and Bitcoin Cash Network can be derived from `NODE_ENV`

## Change Type
- Feature
Added testnet support

- Bug Fix
1. The var `network` is assigned with `settings.NETWORK_BITCOIN` but we are using `NETWORK` as the name in `config/index.js` so its value is always `undefined`. Renamed `NETWORK` as `NETWORK_BITCOIN`
2. Passing `network` as a param in `actions.auth.register()` function

## Testing Steps
1. Run `yarn start:testnet`
2. Should successfully create a wallet

## Code Checklist


packages/blockchain-wallet-v4-frontend/src/config/index.js
```diff
@@ -1,13 +1,15 @@
...
   WALLET_DATA_PATH: 'wallet.data',
   WALLET_PAYLOAD_PATH: 'wallet.payload',
   WALLET_SETTINGS_PATH: 'wallet.settings',
   WALLET_OPTIONS_PATH: 'wallet.options',
   WALLET_KVSTORE_PATH: 'wallet.kvstore',
-  NETWORK: Bitcoin.networks.bitcoin,
+  NETWORK_BITCOIN: Bitcoin.networks[isTestnet ? 'testnet' : 'bitcoin'],
   NETWORK_ETHEREUM: 1,
-  NETWORK_BCH: BitcoinCash.networks.bitcoin
+  NETWORK_BCH: BitcoinCash.networks[isTestnet ? 'testnet' : 'bitcoin']
```
plondon: Small nitpick. We're trying to use currency codes instead of currency names. Can you change to NETWORK_BTC?

packages/blockchain-wallet-v4-frontend/src/config/index.js
```diff
@@ -1,13 +1,15 @@
 import Bitcoin from 'bitcoinjs-lib'
 import BitcoinCash from 'bitcoinforksjs-lib'
 
+const isTestnet = process.NODE_ENV === 'testnet'
```
plondon: Correct is process.env.NODE_ENV, also this doesn't seem to work even if it's correct

packages/blockchain-wallet-v4-frontend/src/config/index.js
plondon: If you hardcode `testnet` for now, I noticed a failure and helpful :) note here: http://github.com/blockchain/blockchain-wallet-v4-frontend/blob/e6b67ab820bb6e262dfb2f89ac3f3389d1c1fcb6/packages/blockchain-wallet-v4/src/types/HDAccount.js#L93

plondon: I also noticed we are hardcoding `Bitcoin.networks.bitcoin` in a few places, which will need to be replaced by settings config

lzhuor: @plondon Thank you! 

1. I will address the naming.
2. Good catch! I caught that TODO as well and has addressed it in local. 
3. Great! I will search for these hardcoding reference

packages/blockchain-wallet-v4-frontend/src/config/index.js
lzhuor: Apologize for my late night stupidness. This will not work because process.env.NODE_ENV will not be known by the client. I will address it.

packages/blockchain-wallet-v4-frontend/src/config/index.js
```diff
@@ -1,13 +1,15 @@
 import Bitcoin from 'bitcoinjs-lib'
 import BitcoinCash from 'bitcoinforksjs-lib'
 
+const isTestnet = false
```
plondon: I think a better way to do this is to determine the network through wallet-options. https://github.com/blockchain/wallet-options/blob/master/prod/wallet-options-v4.json#L30. So I will probably end up removing NETWORK related things from this config

plondon:
@lzhuor thanks for starting this off. I put a bit more work in but there are still a few todos before we can merge it.

Sending works, but only if you override the isValidAddress check here: https://github.com/lzhuor/blockchain-wallet-v4-frontend/blob/fix/testnet_support/packages/blockchain-wallet-v4/src/utils/btc.js#L16. We need to pass the network to this function.

lzhuor:
@plondon Thank you for your patience! I was drenched in my personal stuff in the past week. Looking into it now!

plondon:
@lzhuor apology unnecessary, really appreciate your contribution!

plondon:
Not sure why the build failed but moving this to another PR

PR #833, opened 2018-07-30
feat(Register): added password criteria

jstettner:
## Description
Added more descriptive helper text to the password criteria. #899

screen shot 2018-07-30 at 9 56 50 am
screen shot 2018-07-30 at 9 56 26 am
screen shot 2018-07-30 at 9 56 35 am
screen shot 2018-07-30 at 9 56 43 am

## Change Type
- Feature
- Upgrades

## Code Checklist


schnogz:
Super cool idea. 👍 

I'm wondering if when displaying the estimated crack time, we should use `offline_fast_hashing_1e10_per_second` instead of `offline_slow_hashing_1e4_per_second`..?  This would maybe help push the user into using an even more secure password. Just thinking aloud here.

PR #853, opened 2018-08-09
fix(Tech debt): replaced most .data occurences in remotes

tony-blockchain:
## Description
Replaced `.data` attribute usage on `Remote` objects.
Tested send/receive, settings.
Sfox/coinify and ideally everything else should be double checked.

## Change Type
- Bug Fix

## Code Checklist


schnogz:
Doing Gods work @tony-blockchain  🙌

PR #862, opened 2018-08-13
Fix sms modal closing on incorrect code

prwelber:
## Description
994

## Change Type
- Bug Fix

## Testing Steps

## Code Checklist


packages/blockchain-wallet-v4-frontend/src/modals/MobileNumberVerify/index.js
```diff
@@ -16,9 +16,12 @@ class MobileNumberVerifyContainer extends React.PureComponent {
     this.handleChange = this.handleChange.bind(this)
   }
 
+  componentDidUpdate (prevProps) {
+    if (not(prop('smsVerified', prevProps)) && prop('smsVerified', this.props)) this.props.modalActions.closeAllModals()
+  }
+
```
plondon: I feel like this should be done in a saga and not here. I also sometimes want to do things this way but I think sagas are cleaner.

If verifyMobile settings saga is successful close that modal specifically. Also easier to test. If we can't do that because of sfox we should figure it out still.

PR #863, opened 2018-08-13
Removed deprecated componentWillMount and componentWillReceiveProps

sendmebiscuits:
## Description
Removed deprecated lifecycle methods componentWillMount and componentWillReceiveProps from Carousel/index.js and Form/SelectInput/index.js

## Change Type
Update - Remove deprecated methods


sendmebiscuits:
Ahh, alright no problem. I was thinking just removing those methods was a little weird 😅. I'll make those changes and work on the src files in my next push.

sendmebiscuits:
For the componentDidUpdate methods, I also updated the argument being passed in from nextProps to prevProps as per the React documents. I was also unsure if I should've added getDerivedStateFromProps where setState was used without a conditional. Please let me know if there are any changes you want made!

prwelber:
This is great @sendmebiscuits, thanks. I'm going to close this and open another PR for the issue with some comments where we can continue working on it. 🎉

PR #871, opened 2018-08-15
Fix(Update and pin versions of developer specific NPM dependencies)

schnogz:
## Description
- Fixes issues where a few NPM dependencies where being installed over HTTP
- Removed some unused dependencies and pins the versions of all others
- Upgraded all developer only packages to latest minor and patch level versions

## Change Type
- Upgrades
- Security Fix

## Testing Steps
- App still works
- NPM commands still function properly

## Future TODOs
- We have a growing number of dependencies that are now behind on major version levels.  We should address this soon!

## Code Checklist


package.json
```diff
@@ -121,91 +121,92 @@
     "manage:translations":
       "yarn workspace blockchain-wallet-v4-frontend manage:translations"
   },
-  "dependencies": {},
+  "dependencies": {
+    "compression": "https://registry.npmjs.org/compression/-/compression-1.7.2.tgz",
+    "onetime": "https://registry.npmjs.org/onetime/-/onetime-1.1.0.tgz"
+  },
```
schnogz: These were the two indirect dependencies that were being installed over HTTP. This is sort of a brittle hack but ensures that they get installed over HTTPS via the `yarn.lock` file.

PR #874, opened 2018-08-15
Fix/testnet support

plondon:
## TODO
- [x] Fix kvstore entries
- [x] Fix send
- [x] Fix add wallet account
- [ ] Disable other currencies?

## Description
Pull in network from wallet-options

## Change Type
- Feature
Added testnet support

- Bug Fix
~~1. The var `network` is assigned with `settings.NETWORK_BITCOIN` but we are using `NETWORK` as the name in `config/index.js` so its value is always `undefined`. Renamed `NETWORK` as `NETWORK_BITCOIN`
2. Passing `network` as a param in `actions.auth.register()` function~~

## Testing Steps
1. Change wallet-options-v4 in config/ to use testnet instead of bitcoin network.
2. Run `yarn start:testnet`
3. Should successfully create a wallet

## Code Checklist


plondon:
@lzhuor thanks again for your help, heres the PR

lzhuor:
Thank you @plondon ! Greatly appreciate your help!

packages/blockchain-wallet-v4/src/types/KVStoreEntry.js
```diff
@@ -61,19 +61,19 @@ export const fromKeys = (entryECKey, encKeyBuffer, typeId) => {
   })
```
pernas: I don't remember if kvstore works with testnet addresses. Have you tried that @plondon ?

plondon: I actually need to check this more because if I set a kvstore value (e.g for shapeshift state) then login breaks.

Note to self:
```
at fromObject (index.js:311)
    at from (index.js:137)
    at Function.push.../../node_modules/buffer/index.js.Buffer.from (index.js:149)
    at magicHash (index.js:37)
    at Object.verify (index.js:57)
    at KVStoreEntry.js:136
    at _curryN.js:34
    at _arity.js:22
    at KVStoreEntry.js:160
    at _curryN.js:34
```

schnogz:
Overall looks pretty good. Two feature requests though.
1) Can we change the Webpack dev server wallet options to proxy to return `wallet-options-v4.json` with the network set to testnet. That way we dont have to manually modify the file.
2) Similar to v3, we should have some sort of red warning banner in the header of the app that indicates testnet is in use. Hopefully people wont lose funds then.

packages/blockchain-wallet-v4/src/types/KVStoreEntry.js
plondon: kvstore works now

PR #879, opened 2018-08-16
get guid from url

salomegeo:
## Description
1216

## Change Type
- Feature
- Bug Fix
- CI/CD
- Upgrades
- Other (add explanation)

## Testing Steps

## Code Checklist


packages/blockchain-wallet-v4-frontend/src/scenes/Login/index.js
```diff
@@ -60,12 +60,12 @@ class LoginContainer extends React.PureComponent {
...
     }
 
-    return lastGuid ? (
-      <Login
-        {...this.props}
-        initialValues={{ guid: lastGuid }}
-        {...loginProps}
-      />
+    const path =
+      this.props.location.pathname && this.props.location.pathname.split('/')[2]
```
schnogz: Should we have some sort of isGuid regex check instead of assuming that what you found on pathname is a GUID? example: [blockchain].com/#/login/some-future-route-idk

salomegeo:
I disagree. we don't check validity of guids, you can enter bahblah in the guid field. so it's ok if we enter blockchain.com/#/login/blahblah

I think some-future-route was a solid argument, but can we worry about this when it happens?

schnogz:
That's just a poor UX. 

We have a regex built already to check if a given string looks like a guid. `services/validationHelper/index.js` `isGuid()`, let's use it please.

salomegeo:
fine

PR #894, opened 2018-08-21
do not use getOrFail

prwelber:
## Description
I missed this when I went through the `.data` cleanup. We shouldn't use getOrFail here, because the trade is not always required. For a new account, for example, this was breaking the Coinify experience. QA would have been blocked by this.

## Change Type
- Bug Fix

## Testing Steps

## Code Checklist


plondon:
So let's make this for 4.3.x please if it blocks qa

PR #917, opened 2018-08-24
Fix/imported priv keys

plondon:
## Description
1301

## Change Type
- Bug Fix

## Testing Steps
Import bip38 private key. Let me know if you need one.

## Code Checklist

### DIFF
--- a/packages/blockchain-wallet-v4-frontend/src/data/components/importBtcAddress/sagas.spec.js
+++ b/packages/blockchain-wallet-v4-frontend/src/data/components/importBtcAddress/sagas.spec.js
@@ -1,0 +1,76 @@
+import { testSaga } from 'redux-saga-test-plan'
+import { coreSagasFactory } from 'blockchain-wallet-v4/src'
+import { actions } from 'data'
+import sagas from './sagas.js'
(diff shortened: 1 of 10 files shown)


packages/blockchain-wallet-v4-frontend/src/data/components/importBtcAddress/sagas.spec.js
```diff
@@ -0,0 +1,76 @@
...
+
+describe('importBtcAddress sagas', () => {
+  describe('importBtcAddressSubmitClicked', () => {
+    describe('non bip-38 addr', () => {
+      const saga = testSaga(importBtcAddressSubmitClicked)
+      it('should select form state', () => {
+        // Fails
+        // saga.next().select(selectors.form.getFormValues('importLegacyAddress'))
+        saga.next()
+      })
```
plondon: Do you know why this fails @tony-blockchain

tony-blockchain: It's likely that `selectors.form.getFormValues('importBtcAddress')` fails without form in state
general saga `select` tests don't need the state to be present

PR #925, opened 2018-08-27
Fix/dropdown color

plondon:
## Description
1309
Better dropdowns, theme support

## Change Type
- Upgrades

## Code Checklist

### DIFF
--- a/packages/blockchain-info-components/src/Form/SelectInput/template.js
+++ b/packages/blockchain-info-components/src/Form/SelectInput/template.js
@@ -15,7 +15,7 @@ const StyledSelect = styled(Select)`
     color: ${props => props.theme['gray-5']};
     background-color: ${props => props.theme['white']};
     cursor: pointer;
-    minheight: 40px;
+    min-height: 40px;
     border-radius: 0;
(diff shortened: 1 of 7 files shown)


packages/blockchain-info-components/src/Form/SelectInput/template.js
```diff
@@ -26,15 +28,35 @@ const StyledSelect = styled(Select)`
   }
 `
 
-const customStyles = {
+const customStyles = theme => ({
   control: styles => ({
     ...styles,
-    backgroundColor: 'white',
+    backgroundColor: Color('white', theme),
```
tony-blockchain: could this styles be moved under `StyledSelect` `.control`? connector won't be needed if it's possible

plondon: The docs suggests this approach is best, also they recently merged this [PR](https://github.com/JedWatson/react-select/pull/2839) which will allow us to completely control the theme, but has not been released yet.

I don't think we'd need any css-in-js, which I think is less maintainable anyway.

plondon: Done

PR #935, opened 2018-08-29
Fixed a few infinite loops

sendmebiscuits:
## Description
Fixed infinite loops based on suggestions. 

## Change Type
- Feature
- Bug Fix
- CI/CD
- Upgrades
- Other (add explanation)

## Testing Steps

## Code Checklist

### DIFF
--- a/packages/blockchain-info-components/src/Form/SelectInput/index.js
+++ b/packages/blockchain-info-components/src/Form/SelectInput/index.js
@@ -24,7 +24,7 @@ class SelectInputContainer extends React.PureComponent {
 
   componentDidUpdate (prevProps) {
     /* eslint-disable */
-    if (!equals(this.props.value, prevProps.value)) {
+    if (!equals(this.props.value, this.state.value)) {
       this.setState(
         {
           value:
--- a/packages/blockchain-info-components/src/Form/SelectInput/index.js
+++ b/packages/blockchain-info-components/src/Form/SelectInput/index.js
@@ -22,7 +22,7 @@ class SelectInputContainer extends React.PureComponent {
     this.handleKeyDown = this.handleKeyDown.bind(this)
   }
 
-  componentDidUpdate (prevProps) {
+  componentDidUpdate () {
     /* eslint-disable */
     if (!equals(this.props.value, this.state.value)) {
       this.setState(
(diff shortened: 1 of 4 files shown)


packages/blockchain-wallet-v4-frontend/src/modals/CoinifyExchangeData/Create/AcceptTerms/index.js
```diff
@@ -18,7 +18,7 @@ class AcceptTermsContainer extends Component {
   }
 
   componentDidUpdate (prevProps) {
-    if (prevProps.signupError) {
+    if (prevProps.signupError||this.state.busy) {
```
tony-blockchain: it should be `&&` here, otherwise it would still trigger when `this.state.busy` is falsy and `prevProps.signupError` is truthy.
Also spaces around operator are missing

packages/blockchain-info-components/src/Form/SelectInput/index.js
```diff
@@ -24,7 +24,7 @@ class SelectInputContainer extends React.PureComponent {
 
   componentDidUpdate (prevProps) {
```
tony-blockchain: prevProps is unneeded param in this case

packages/blockchain-wallet-v4-frontend/src/modals/CoinifyExchangeData/Create/AcceptTerms/index.js
sendmebiscuits: Whoops, got it.

plondon:
@prwelber what's the status of all this?

packages/blockchain-wallet-v4-frontend/src/modals/CoinifyExchangeData/Create/AcceptTerms/index.js
```diff
@@ -18,7 +18,7 @@ class AcceptTermsContainer extends Component {
   }
 
   componentDidUpdate (prevProps) {
-    if (prevProps.signupError) {
+    if (prevProps.signupError && this.state.busy) {
```
prwelber: @sendmebiscuits I think we should check `this.props.signupError` instead of `prevProps.signupError`otherwise this `if` block will never run. You also need to use eslint-disable and eslint-enable or the linter will complain about setState inside componentDidUpdate. In this case, there shouldn't be any updating loop, I checked.

A good way to test this is to set an incorrect` partnerId` in `config/wallet-options` and then try to create a coinify account. You should see the scenario play out. Looks good otherwise!

sendmebiscuits: Ah alright that makes sense! I'm also gonna add the eslint-disable/enable to the other componentDidUpdate and componentDidMount methods in the other files for consistency's sake.

prwelber: Thanks, sounds good to me 👍

PR #955, opened 2018-09-05
Fix/bch export

prwelber:
## Description
1307

## Change Type
- Bug Fix

## Testing Steps

## Code Checklist

### DIFF
--- a/packages/blockchain-wallet-v4/src/redux/data/bch/sagas.js
+++ b/packages/blockchain-wallet-v4/src/redux/data/bch/sagas.js
@@ -83,9 +83,9 @@ export default ({ api }) => {
     const { address, start, end } = payload
     try {
       yield put(A.fetchTransactionHistoryLoading())
-      const convertedAddress = fromCashAddr(address)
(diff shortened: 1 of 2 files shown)


prwelber:
TODO

- [ ] fix issue where selecting a BCH imported address in the tx list returns an error from `/multiaddr`

plondon:
Was this ever working @TheLeoB?

If yes would rather fix the root cause than patch a fix.

TheLeoB:
@plondon I feel like this is something we could have caught with testing, but I can't say for sure. Is there a way for us to look at builds older than 4.2.1?

plondon:
release/v4.1.x branch is still available

TheLeoB:
That branch runs 4.2.1

plondon:
Here is the commit hash for 4.0.6 87ac43f8488b37c345f46479497e73d1b9ca0162

So `git co master` then `git reset 87ac43f8488b37c345f46479497e73d1b9ca0162` should bring you back there, then reset master to the latest

PR #1069, opened 2018-10-25
fix(BCH): Check if bch change index got behind btc by gap limit

plondon:


packages/blockchain-wallet-v4-frontend/src/data/modules/addressesBch/sagas.js
```diff
@@ -24,7 +28,54 @@ export default ({ coreSagas }) => {
...
+      index,
+      bchChangeIndex,
+      btcChangeIndex,
+      state
+    )
+
+    // if there are addrs to display show them else carry on
+    /* eslint-disable */
+    if (addrs.length) {
+      window.alert(
```
tony-blockchain: nice solution although these might be blocked by browser, console.log is safer imo

plondon: We're a bit worried users won't be able to easily open the console. I'll check in with support after the fix is deployed, if anyone has issues we'll switch to console log.

packages/blockchain-wallet-v4-frontend/src/scenes/Settings/Addresses/Bch/Wallets/template.success.js
```diff
@@ -160,7 +161,16 @@ const Success = props => {
                     defaultMessage='Show xPub'
                   />
                 </ClickableText>
-              )
+              ),
+              <ClickableText
+                size='small'
+                onClick={() => onShowChangeAddrs(wallet.value)}
```
tony-blockchain: onClick={onShowChangeAddrs.bind(null, wallet.value)}

packages/blockchain-wallet-v4-frontend/src/scenes/Settings/Addresses/Bch/Wallets/template.success.js
```diff
@@ -160,7 +161,16 @@ const Success = props => {
                     defaultMessage='Show xPub'
                   />
                 </ClickableText>
-              )
+              ),
+              <ClickableText
```
tony-blockchain: I think we should use selector to determine the difference between btc and bch change addresses and only show link in this case

PR #1098, opened 2018-11-07
fix(upgradetohd): add missing arg to curried fn

merildev:
## Description
Upgrade non HD wallet was broken because of a missing arg in a curried function.

## Change Type
- Bug Fix

## Testing Steps
N/A

## Code Checklist


pernas:
Amazing @merildev . This bug was probably introduced when the testnet support was added and the signature of all those functions changed to accept another parameter and that one was left unchanged. A good example why having type checking is important. That bug has been out there for months completely unnoticed.

PR #1104, opened 2018-11-07
Feat(Lockbox app manager)

schnogz:
## Description
- Adds an application manager modal that is shared between the settings page and the new device onboarding
- Fixes an issue where we were not correctly parsing custom providerIds in network responses from Ledger.  We will now get the correct apps and versions on network calls for both device types.
- Removes and cleans up all the nasty bulk app install related code and redux states
- No longer require either a `deviceType` or `deviceIndex` in `pollForDeviceApp` saga.  It will pull that info from previous connection if needed.
- Added unit tests
- No idea what all the "No changes" to all those files are with `100644 → 100755`. Perhaps from switching repos on the branch? Shouldn't hurt anything either way...

## TODOs
- Need to verify correct status codes are returned during different install and uninstall app situations. Do this once latest versions of apps are pushed by Ledger
- Need to refine some of the styling once wireframes are finished @plondon 😏 

## Change Type
- Feature

## Testing Steps
- Apps should be able to individually install and uninstall on both Ledger and Blockchain devices.

## Code Checklist


packages/blockchain-wallet-v4-frontend/package.json
```diff
@@ -36,9 +36,10 @@
     "collectCoverageFrom": [
       "src/**/*.js",
       "!src/**/rootSaga.js",
+      "!src/data/components/sagas.js",
```
tony-blockchain: middleware & modules should also be ignored

schnogz: fixed 👍

packages/blockchain-wallet-v4-frontend/src/modals/Lockbox/AppManager/index.js
```diff
@@ -164,7 +162,7 @@ class AppManagerContainer extends React.PureComponent {
     })
     const appListView = appVersionInfos.cata({
       Success: apps => {
-        return apps.map((app, i) => {
+        const appList = apps.map((app, i) => {
           const coin = getKeyByValue(app.name)
           return (
             <App
```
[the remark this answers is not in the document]
schnogz: fixed 👍

PR #1179, opened 2018-11-20
Refactor(Announcement component)

schnogz:
## Description
- Modest refactor of the announcements into a single place. Could be refactored more but calling it for now
- Adds testnet warning to both public and wallet layout

## Change Type
- Bug Fix
- Refactor

## Screenshot
screen shot 2018-11-20 at 4 26 54 pm

## Code Checklist


schnogz:
Just realized we dont have support to show multiple announcements from wallet options at the same time.

Adding now.

packages/blockchain-wallet-v4-frontend/src/components/Announcements/Service/index.js
```diff
@@ -24,15 +21,31 @@ class ServiceAnnouncement extends React.PureComponent {
...
-          <Announcement
-            announcement={val.announcements[alertArea]}
-            language={val.language}
-            collapsed={val.collapsed}
-            handleDismiss={this.handleDismiss}
-            toggleCollapse={this.toggleCollapse}
-          />
+        return val.showTestnetWarning || val.showAnnounce ? (
+          <React.Fragment>
+            {val.showTestnetWarning && (
```
plondon: Why can't testnet warning come from wallet-options-v4.json for only testnet?

plondon: I see that makes sense, however for the ~1year we had testnet running we only showed this one banner. Announcement banners in general should be used sparingly and only in cases people could lose funds. Considering testnet is all fake funds I think one main public/wallet banner that says these funds are not real should suffice.

PR #1253, opened 2018-12-05
Feat(Service announcements fix & redux-persist v5 migration)

schnogz:
## Description
- Fix to ensure service announcements show even when there is cache miss for announcement states (e.g. incognito browsing)
- Easiest fix was to update to v5 of `redux-persist` that adds the `PersistGate` component that will not render the UI until the cache is completely loaded. Without this there would be screen flickering of the banner if it was previously dismissed
- Followed the v4 to v5 migration guide [here](https://github.com/rt2zz/redux-persist/blob/master/docs/MigrationGuide-v5.md#experimental-v4-to-v5-state-migration)

## Change Type
- Bug Fix
- Upgrades

## Testing Steps
- Ensure your wallet GUID is still being auto-loaded from cache upon signon

## Code Checklist

### DIFF
--- a/packages/blockchain-wallet-v4-frontend/package.json
+++ b/packages/blockchain-wallet-v4-frontend/package.json
@@ -136,7 +136,7 @@
     "redux": "3.6.0",
     "redux-form": "7.3.0",
     "redux-logger": "3.0.6",
-    "redux-persist": "4.8.3",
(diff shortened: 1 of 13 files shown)


plondon:
Should this be Next Release?

PR #1274, opened 2018-12-11
Fix(Lockbox connection polling upgrades)

schnogz:
## Description
Addresses a long standing issue in which if a modal was closed while we were polling for a device connection or interacting on the device in anyway, that polling/running saga would never stop and would be left in a broken state. This caused issue with any subsequent device interactions because previous polling was usually still occurring.

## Changes
- Disables Lockbox modals from being closed via `onClickOutside`. All modals need to be explicitly closed by user to ensure we cancel currently running sagas and connection polling properly
- Adds modal headers and close icons to all Lockbox modals
- `initializeNewDeviceSetup`, `updateDeviceFirmware` & `pollForDeviceApp` sagas are now setup as `races` between its own completion and the modal close event.  `finally` blocks are added to each and will clean up its own saga state as needed
- Adds consistent naming to Lockbox modals and its folder structure
- Re-enables Lockbox connection polling & socket messaging console logging

## Change Type
- Bug Fix

## Testing Steps
- Switch between different modals that poll for device connections or have multi-step flows.  Close modals half way between and you should see previous saga/polling cancelled and your new modal connections should function.

## Code Checklist


packages/blockchain-wallet-v4-frontend/src/services/LockboxService/utils.js
```diff
@@ -83,16 +83,16 @@ const createDeviceSocket = (transport, url) => {
...
 
     ws.onerror = e => {
-      //console.info('ERROR', { message: e.message, stack: e.stack })
+      console.info('ERROR', { message: e.message, stack: e.stack })
       o.error(e.message, { url })
     }
 
     ws.onclose = () => {
-      //console.info('CLOSE')
+      console.info('CLOSE')
```
tony-blockchain: are these left uncommented on purpose?

schnogz: We had them uncommented for a very long time for debugging purposes in higher envs. I just re-enabled them as we have a revamp of a few Lockbox features coming up.  

Will definitely disable again though in the future.

tony-blockchain: yeah seen them working when tested lockbox

PR #1293, opened 2018-12-17
Matomo

plondon:


packages/blockchain-wallet-v4-frontend/src/data/analytics/reducers.js
```diff
@@ -0,0 +1,18 @@
...
+import { insert } from 'ramda'
+
+const INITIAL_STATE = []
+
+const analytics = (state = INITIAL_STATE, action) => {
+  const { type, payload } = action
+
+  switch (type) {
+    case AT.LOG_EVENT: {
+      return insert(0, payload, state)
```
plondon: Can you add some error handling here

If event is not in ANALYTICS throw error
If payload is incorrect (not an array, check length?)

Another way would be to construct the payload yourself, similar to ```const createLog = (msgType, payload) => {
    return {
      type: msgType,
      file: payload.file,
      method: payload.method,
      timestamp: Date.now(),
      message: JSON.stringify(payload.message)
    }
  }```

prwelber: I thought `assert` worked well here. Really the two most likely failures are empty data and data being passed that is inconsistent with our "schema". In the future we can also check the event `category` so we're not creating duplicates by doing things like passing `send` versus `Send`

packages/blockchain-wallet-v4-frontend/src/data/analytics/reducers.js
```diff
@@ -8,6 +10,10 @@ const analytics = (state = INITIAL_STATE, action) => {
 
   switch (type) {
     case AT.LOG_EVENT: {
+      assert(path(['trackingData', 'length'], payload), 'trackingData cannot be empty')
+      const trackingAction = path(['trackingData', 1], payload)
+      assert(TRACKING_ACTIONS.indexOf(trackingAction) > -1, `tracking action: ${trackingAction} is not recognized`)
```
plondon: Thanks, I prefer contains over indexOf

prwelber: Yeah, much nicer. 

Small note about `contains`: I took a quick peek at the docs and even though we use `contains` throughout the wallet, we should switch to `includes`. Looks like `contains` was deprecated since v0.26.0 (we're on 0.25.0). Anyway, change made.

https://ramdajs.com/docs/#contains
https://github.com/ramda/ramda/issues/2702

plondon:
Don't merge until discussing actions w/ Matomo

packages/blockchain-wallet-v4-frontend/src/providers/AnalyticsTracker/index.js
```diff
@@ -0,0 +1,62 @@
...
+  height: 0px;
+  width: 0px;
+`
+
+class AnalyticsTracker extends React.PureComponent {
+  componentDidUpdate (prevProps) {
+    const { analytics, location, domains } = this.props
+    const { analytics: prevAnalytics, location: prevLocation } = prevProps
+    const targetOrigin = domains.walletHelper
+    const node = ReactDOM.findDOMNode(this)
```
tony-blockchain: please use `React.createRef()` if possible

packages/blockchain-wallet-v4-frontend/src/data/analytics/reducers.js
```diff
@@ -0,0 +1,41 @@
...
+  swapTiming: null,
+  events: []
+}
+
+const analytics = (state = INITIAL_STATE, action) => {
+  const { type, payload } = action
+
+  switch (type) {
+    case AT.LOG_EVENT: {
+      assert(path(['trackingData', 'length'], payload), 'trackingData cannot be empty')
```
tony-blockchain: `assert` module will throw in case of failure, right? Don't think it's good idea to throw error inside reducer

tony-blockchain: also why adding a whole extra module to the production?

plondon: +1 im also not a fan of assert, need to think too much, the whole rest of the code base throws Errors

prwelber: Sure, there are other ways to handle malformed payloads we can use here. BTW- assert is already used in the core, so it wasn't added just for this case.

tony-blockchain: nice, core usage is cool, reminds me that sometimes developers install postgres to use it as csv :D

packages/blockchain-wallet-v4-frontend/src/providers/AnalyticsTracker/index.js
```diff
@@ -44,6 +49,7 @@ class AnalyticsTracker extends React.PureComponent {
       <Iframe
         id='matomo-iframe'
         src={walletHelper + '/wallet-helper/matomo/#/'}
+        innerRef={this.iframe}
```
prwelber: `innerRef` is deprecated in styled-components v4 btw
https://www.styled-components.com/docs/api#innerref-prop

PR #1303, opened 2018-12-18
Fix/tier states

plondon:
## Description
1944, 1927

## Testing Steps
Test a rejected user

## Code Checklist


packages/blockchain-wallet-v4-frontend/src/scenes/Settings/Profile/Menu/index.js
```diff
@@ -0,0 +1,53 @@
...
+  border-bottom: 1px solid ${props => props.theme['gray-1']};
+  a {
+    color: ${props => props.theme['brand-secondary']};
+    text-decoration: none;
+  }
+`
+
+export const Menu = ({ userTiers }) => {
+  const isRejected = all(propEq('state', 'rejected'), userTiers)
+  return isRejected ? (
```
tony-blockchain: if (isRejected) return null or <div /> and then return template would look nice

PR #1322, opened 2018-12-22
Feat(Fallback to home for hits on unregistered routes if user is authenticated)

schnogz:
## Description
- If a user is authenticated and hits an unregistered route, redirect to home page instead of login

## Change Type
Feature

## Testing Steps

## Code Checklist


schnogz:
Yes, I tested both scenarios (authed and unauthed) and it works just fine.

PR #1340, opened 2019-01-02
fix(Forms): use model so form names dont get out of sync

prwelber:
## Description
The form name for the transactions search/filter component was not consistent across the app.

## Change Type
- Bug Fix

## Testing Steps
- see BTC and BCH web sockets now properly work when updating tx feed
- tx feed form should work

## Code Checklist

### DIFF
--- a/packages/blockchain-wallet-v4-frontend/src/data/components/bsvTransactions/sagas.js
+++ b/packages/blockchain-wallet-v4-frontend/src/data/components/bsvTransactions/sagas.js
@@ -1,8 +1,9 @@
 import { select, put } from 'redux-saga/effects'
 import { equals, path, prop } from 'ramda'
-import { actions, selectors } from 'data'
+import { actions, selectors, model } from 'data'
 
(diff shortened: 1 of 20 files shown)


packages/blockchain-wallet-v4-frontend/src/data/components/bsvTransactions/sagas.js
```diff
@@ -1,8 +1,9 @@
 import { select, put } from 'redux-saga/effects'
 import { equals, path, prop } from 'ramda'
-import { actions, selectors } from 'data'
+import { actions, selectors, model } from 'data'
 
 export default () => {
```
prwelber: @schnogz Are these necessary? Is there a transaction feed style component for BSV?

plondon: Yes

plondon:
Thanks, I did this in another branch but using model is better

PR #1421, opened 2019-02-08
feat(header/footer): links rename 

blockfred:
## Description
Navigation/Footer: Update links for markets relaunch,
Data - Markets (links to "/markets") => Data - Prices (links to "/prices")
Products - Principal Strategies (links to "bps.blockchain.com") => Products - Markets (links to "/markets")

## Change Type
- Feature

## Testing Steps
Open verify header/footer links work

## Code Checklist


plondon:
Has https://github.com/blockchain/components been updated? Maybe we should just use that library for markets instead?

PR #1488, opened 2019-02-28
Fix/login error

tony-blockchain:
## Description

## Change Type
- Feature
- Bug Fix
- CI/CD
- Upgrades
- Other (add explanation)

## Testing Steps

## Code Checklist


plondon:
It would be nice to have a `safeParse` function or something here instead

tony-blockchain:
this one in fact is logical error, not connected to parsing - we are allowing objects through where only strings are processed

PR #1500, opened 2019-03-06
Refactor/seed protection

Empowerful:
# Introduction

This PR is both a proof of concept and first step toward protecting the backup phrase.  It is presented here for review but is not intended to be merged (see below for why).

To protect the backup phrase I divided the Web Wallet into two separate applications: one trusted more (the "Root Document") and the other trusted less (the "Main Process").  The Main Process is an `<iframe>` within the Root Document that takes up the whole viewport and is served locally via port 8082.  The two applications communicate with each other via `postMessage`.

To save development time I created each application from a complete copy of the current Web Wallet and then made modifications to them.  The two applications are mostly the same but over time will become more differentiated as I remove code that's no longer relevant to each application.

# Notes

* In the code you'll see the applications sometimes referred to as "realms".  A _realm_ is a JavaScript environment with a distinct namespace.
* The `main-process` directory is not within `packages` because Yarn doesn't support nested workspaces.
* My new inter-realm communication library is in its own subdirectory (`web-microkernel`) because it will be moved into its own repository in the future.

# Use

Install new dependencies:

```shell
$ yarn
```

Start the Main Process on port 8082:

```shell
$ cp config/env/* main-process/config/env
$ cd main-process
$ yarn
$ yarn start:staging
```

Once the Main Process has started, open a second terminal in the top level of the repository and start the Root Document application on port 8080:

```shell
$ yarn start:staging
```

Questions and feedback are highly encouraged.


packages/web-microkernel/src/RealmConnection.js
```diff
@@ -0,0 +1,400 @@
+import * as _ from './lodash-es/lodash.js'
```
tony-blockchain: would be nice to tree-shake other lodash functions before we release

packages/web-microkernel/src/RealmConnection.js
```diff
@@ -0,0 +1,400 @@
...
+
+// Polyfill the behavior of addEventListener(type, listener, { once: true }))
+// because for an unknown reason the above isn't working in Chrome 72.
+const addEventListenerOnce = (target, type, listener) => {
```
tony-blockchain: quite useful stuff to have as a separate util

main-process/packages/blockchain-wallet-v4-frontend/package.json
```diff
@@ -88,6 +88,7 @@
     "@ledgerhq/hw-app-btc": "4.30.0",
     "@ledgerhq/hw-app-str": "4.26.0-beta.ebeb3540",
     "@ledgerhq/hw-transport-u2f": "4.31.0",
+    "@nodeguy/channel": "0.6.4",
```
guy-blockchain: In the interest of reducing external dependencies, would it be better for @blockchain to fork this or create an internal service.

packages/web-microkernel/src/RealmConnection.js
```diff
@@ -0,0 +1,400 @@
...
+    decoder: (context, codes) =>
+      codes.map(code => decodeFromDictionary(context, code))
+  }
+)
+
+// error
+
+const isError = value => value instanceof Error
+
+const encodeError = (context, error) => ({
```
tony-blockchain: would be nice to have resulting types annotated for example in jsdoc, otherwise it's quite difficult to image results of encodings

main-process/packages/blockchain-wallet-v4-frontend/package.json
tony-blockchain: agreed, I've once encountered a problem where personal namespaces drifted off from company needs. Double the support prs but ownership is important

packages/web-microkernel/src/RealmConnection.js
Empowerful: We're not going to use Lodash—it's temporary filler.

main-process/packages/blockchain-wallet-v4-frontend/package.json
Empowerful: @guy-blockchain We currently have 387 production dependencies.  I think it will be an interesting conversation in the future to explore how to minimize dependencies but for now it's not a big deal to add 2 more.

@tony-blockchain I don't understand your point.  We don't own most of the 387 dependencies (e.g., React).  How is this one any different?

packages/web-microkernel/src/RealmConnection.js
Empowerful: The encoding is an implementation detail, not part of the API, and could change over time so I don't think it's a good idea to put it into documentation.

Empowerful:
@tony-blockchain asked to see some examples of the inter-realm encoding.

Here's the handshake of the two realms exchanging their exports with each other:
```JSON
{
  "0": ["object", [["axios", 1], ["dispatch", 2]]],
  "1": ["function", { "key": "13mjn49", "length": 1 }],
  "2": ["function", { "key": "1e1vcyr", "length": 1 }]
}

{
  "0": ["object", [["dispatch", 1]]],
  "1": ["function", { "key": "jh43f1", "length": 1 }]
}
```

Notice the keys are randomly generated so that a compromised realm couldn't guess the keys in the other.

Here's an example of a function application (call).  This is calling the Redux `dispatch` function in the other realm to forward the `AUTHENTICATE` action:

```JSON
{
  "0": [
    "functionApplication",
    {
      "args": {
        "0": ["array", [1]],
        "1": ["object", [["type", 2]]],
        "2": ["string", "AUTHENTICATE"]
      },
      "functionKey": "1e1vcyr",
      "returnValueKey": "1bb1tve"
    }
  ]
}
```

Here's the result of the function application being returned:

```JSON
{
  "0": [
    "functionReturn",
    { "returnValueKey": "1bb1tve", "value": { "0": ["number", 0] } }
  ]
}
```

jakubusi:
Haha

yarn.lock
```diff
@@ -17583,11 +17634,11 @@ uglify-js@3.4.x, uglify-js@^3.1.4:
...
-  integrity sha512-KeQesOpPiZNgVwJj8Ge3P4JYbQHUdZzpx6Fahy6eKAYRSV4zhVmLXoC+JtOeYxcHCHTve8RG1ZGdTvpeOUM26Q==
+  version "3.6.4"
+  resolved "https://registry.yarnpkg.com/uglify-js/-/uglify-js-3.6.4.tgz#88cc880c6ed5cf9868fdfa0760654e7bed463f1d"
+  integrity sha512-9Yc2i881pF4BPGhjteCXQNaXx1DCwm3dtOyBaG2hitHjLWOczw/ki8vD1bqyT3u6K0Ms/FpCShkmfg+FtlOfYA==
   dependencies:
-    commander "~2.20.0"
+    commander "~2.20.3"
     source-map "~0.6.1"
 
 uglifyjs-webpack-plugin@2.1.3:
```
laothuong: 9Yc2i881pF4BPGhjteCXQNaXx1DCwm3dtOyBaG2hitHjLWOczw

Milan-Popich:
Quick and to the point. Is it really necessary for all of these blockchain wannabe development emails coming out? For over four years you failed to complete one key aspect which is the synchronization process not only with the network but also with the customer and the wallet. I should not have wasted five years of my life with this wannabe programming with your failures and your “I deserve participation points” attitude. In business this is beyond unacceptable, you showing that your programming staff is incapable and incompetent of handling the basis of their job duties should have been replaced years ago. Also send me my blockchain API, and find someone that will be capable of completing your basic job duties.

In Health,
Milan Popich

Sent with [ProtonMail](https://protonmail.com) Secure Email.

‐‐‐‐‐‐‐ Original Message ‐‐‐‐‐‐‐
Op zondag, oktober 27, 2019 10:13 AM, laothuong <notifications@github.com> schreef:

scripts/mock-service-wallet.js
```diff
@@ -252,41 +255,39 @@ const { version } = require(`../package.json`)
...
-                domains,
-                request: R.pick([`headers`, `hostname`, `url`], request)
-              }
-            })
-          })
-
-          router(request, response)
-          proxy.web(request, response, { target: mainProcessUrl })
+      const mainRouter = express.Router()
+
```
JohnAckison: Thanks

PR #1552, opened 2019-03-26
fix: removed autofocus from add wallet so errorborder doesnt show rig…

TheLeoB:
## Description
1312

## Change Type
- Bug Fix

## Testing Steps

## Code Checklist


plondon:
Autofocus feature is nice. I think instead of taking it away we should fix the validation happening too soon.

PR #1561, opened 2019-04-01
fix(Swap): use mempool fees from provisional payment

plondon:
## Description
https://blockchain.atlassian.net/browse/WEB4-2315
@tony-blockchain a bit simpler than what we discussed. Not checking if fee is in some range, just simply using fee from provisional payment. I think this code is less offensive

## Change Type
- Bug Fix

## Testing Steps

## Code Checklist


packages/blockchain-wallet-v4-frontend/src/data/components/exchange/exchange.sagas.js
```diff
@@ -320,6 +320,7 @@ export default ({ api, coreSagas, networks }) => {
       yield put(
         A.setSourceFee({
```
tony-blockchain: setSourceFee is done 3 times, I guess we should update mempoolFees with some value everywhere

PR #1579, opened 2019-05-01
Comment out failing tests for now

plondon:


plondon:
@tony-blockchain could you have a look when you have a chance...

PR #1607, opened 2019-05-08
feat(Matomo): if user id, set for matomo

plondon:
@schnogz does this make sense? Request from data team


plondon:
Sure thing

plondon:
Tried to fix but yield take does not return the selector it returns the action.

PR #1610, opened 2019-05-09
Chore/release it

plondon:
Moved from @webpro's PR to here

## Description

Update release-it + config to v11.

Note that the `changelog` executable in the `git._changelog` option is not there. Not sure what package is needed for this. You can fix this and remove the underscore from `_changelog`.

## Change Type

Releases

## Testing Steps

`npm run release`

## Code Checklist


.release-it.json
```diff
@@ -1,20 +1,23 @@
...
+    ]
   },
   "git": {
+    "requireCleanWorkingDir": false,
+    "requireUpstream": false,
+    "_changelog": "changelog -t $(git describe --abbrev=0 --tags) -f -",
+    "commitMessage": "chore(Release): v${version}",
+    "tagName": "v${version}",
+    "tagAnnotation": "Release v${version}",
     "pushArgs": "--no-verify"
```
plondon: @webpro just wanted to let you know --no-verify didn't work before. Now it does! Thanks!

.release-it.json
```diff
@@ -1,20 +1,23 @@
...
   },
   "github": {
     "release": true,
     "preRelease": true,
```
plondon: @webpro I don't think preRelease is working, or maybe the flag was changed? It creates the `Latest Release` tag, but these are all pre releases, I change to `Latest Release` when we deploy to prod.

PR #1619, opened 2019-05-13
fix(XLM): do not clear send form on back

plondon:
@schnogz do you remember why we were doing this in the first place?


schnogz:
i do not recall... should have added a comment 😬

PR #1697, opened 2019-06-27
refactor(migration): Remove unused domain migration code.

Empowerful:
## Description
@plondon:
> I think for this code what happened was we were preparing for .info .com move, and if v4 was
deployed before the migration we were going to need this redirect.  But we deployed v4 after, so
this code was preliminary and is no longer necessary.

**After this PR lands in production we should inform the backend team that we no longer consume the `enableDomainMigrationRedirects` option.**

## Change Type
- Other (add explanation)
Refactor

## Testing Steps
No test.

## Code Checklist

### DIFF
--- a/packages/blockchain-wallet-v4-frontend/src/layouts/Public/index.js
+++ b/packages/blockchain-wallet-v4-frontend/src/layouts/Public/index.js
@@ -10,13 +10,6 @@ import Alerts from 'components/Alerts'
 import ErrorBoundary from 'providers/ErrorBoundaryProvider'
 import { selectors } from 'data'
 import media from 'services/ResponsiveService'
-import { isOnDotInfo } from 'services/MigrationService'
-
-const defaultDomains = {
-  root: 'https://blockchain.info',
-  comWalletApp: 'https://login.blockchain.com',
-  comRoot: 'https://blockchain.com'
-}
 
 const Wrapper = styled.div`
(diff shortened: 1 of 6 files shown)


plondon:
Thanks. Also:

https://github.com/blockchain/blockchain-wallet-v4-frontend/blob/development/packages/blockchain-wallet-v4/src/redux/walletOptions/__mocks__/wallet-options-v4.js#L6

https://github.com/blockchain/blockchain-wallet-v4-frontend/blob/development/config/mocks/wallet-options-v4.json#L6

packages/blockchain-wallet-v4-frontend/src/layouts/Public/index.js
```diff
@@ -102,9 +81,7 @@ class PublicLayoutContainer extends React.PureComponent {
 }
 
 const mapStateToProps = state => ({
-  pathname: selectors.router.getPathname(state),
-  domainsR: selectors.core.walletOptions.getDomains(state),
-  migrationRedirectsR: selectors.core.walletOptions.getMigrationRedirects(state)
+  pathname: selectors.router.getPathname(state)
```
plondon: Is pathname still used?

Empowerful: No it's not; nice catch.

PR #1744, opened 2019-07-23
fix(LinkContext): create full deep link with required parameters

jtormey:
## Description
Just adding the param to the deep link directly doesn't seem to work, this creates a longer link but one that should be able to pass on the `context` param.

@bchrisarriola does this seem more like what you need?

## Change Type
- Feature

## Code Checklist

### DIFF
--- a/packages/blockchain-wallet-v4-frontend/src/scenes/VerifyEmailToken/index.js
+++ b/packages/blockchain-wallet-v4-frontend/src/scenes/VerifyEmailToken/index.js
@@ -9,6 +9,9 @@ import Error from './template.error'
 import { Wrapper } from 'components/Public'
 
 const VALID_CONTEXTS = ['PIT_SIGNUP', 'KYC', 'SETTINGS']
+const PARAM_DEEP_LINK_PATH = 'email_verified'
+const PARAM_ISI = '493253309'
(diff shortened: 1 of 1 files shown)


packages/blockchain-wallet-v4-frontend/src/scenes/VerifyEmailToken/index.js
```diff
@@ -27,12 +30,21 @@ class VerifyEmailToken extends React.PureComponent {
...
-      ? 'https://blockchain.page.link/email_verified'
-      : 'https://blockchainwalletstaging.page.link/email_verified'
+      ? 'https://blockchain.page.link/'
+      : 'https://blockchainwalletstaging.page.link/'
 
-    const isValidContext = VALID_CONTEXTS.indexOf(context) > -1
+    const params = new URLSearchParams()
+    params.set('deep_link_path', PARAM_DEEP_LINK_PATH)
+    params.set('isi', PARAM_ISI)
+    params.set('ibi', PARAM_IBI)
```
bchrisarriola: One more param that needs to be added as well:

```
apn=piuk.blockchain.android
```

jtormey: Added, thanks!

PR #1745, opened 2019-07-23
fix(pit): copy changes for pit landing page

plondon:
2699 cc @jjBlockchain


plondon:
Screen Shot 2019-07-23 at 3 16 32 PM

Looks good! Can you have the image stack on mobile below the title?

PR #1748, opened 2019-07-24
fix(pit-linked-login): changed color, font-weight, and added margin

jjBlockchain:
## Ticket Number
WEB4-2701


jjBlockchain:
green500 has been added to DarkMode.js

PR #1762, opened 2019-07-30
feat(bitpay): implement BitPay's json payment protocol 

will-blockchain:
## Description (optional)
Support opening BitPay invoices in the web wallet through bitcoin linking. 

## Testing Steps (optional)
1. Login to bitpay.com (slack me I can give you a login)
2. Go to https://bitpay.com/dashboard/payment-tools/payment-buttons
3. Scroll to the bottom and click the Pay with BitPay button to generate a new BitPay invoice.
4. When the invoice opens, select Bitcoin, then click the QR code that pops up to copy the bitcoin:?r=... link. (Their pay in wallet button functionality is broken, have to manually copy bitcoin link)
5. Paste this bitcoin:?r=... link into a new tab, wait for redirect, login when prompted, and wait for send modal popup. 
6. (you must also have Bitcoin link handling enabled in settings)

Testing details:
All inputs are disabled except for the description detailing the invoice. This lets users optionally include more details to specify what sort of thing they are buying from the merchant. 
(I've enabled the amount field for now as there is a bug in production @schnogz said he would check out where the amount and description are not being passed to the second step confirm screen)
A countdown timer is included above the To field to give the user feedback on how long they have to pay the invoice. 

Tests:
1. In order to bypass the amount bug, delete and retype the exact Bitcoin amount in the amount field, click continue to go to the second step confirm screen and complete the payment, BitPay invoice will reflect payment promptly.
2. Wait for the invoice to expire which will cause the send modal to close and a BitPay invoice expired info modal to popup informing the user that the invoice has expired.


packages/blockchain-wallet-v4-frontend/src/components/Form/CountdownTimer/template.js
```diff
@@ -19,13 +19,27 @@ const Wrapper = styled.div`
...
 const CountdownTimer = props => (
   <Wrapper {...props}>
-    <Text size='11px' weight={400} color='gray-3'>
-      <FormattedMessage
-        id='scenes.exchange.secondstep.expiry_text'
-        defaultMessage='This quote will refresh in: {timeLeft}'
-        values={{ timeLeft: props.timeLeft }}
-      />
-    </Text>
+    {!props.payProInvoice ? (
```
plondon: Instead of this (and instead of hardcoding the text 'This quote...'). It would be good to require this component take some type of property like countdownMessage (?) and pass that in from the parent component that uses CountdownTimer:

PARENT:
```
const countdownMessage = <Text
        size='14px'
        weight={500}
        color={getCountdownColor(props.expiryDate)}
      >
        <FormattedMessage
          id='modals.sendbtc.firststep.paypro_expiry_text'
          defaultMessage='Remaining time {timeLeft}'
          values={{ timeLeft: props.timeLeft }}
        />
      </Text>

...

<CountdownTimer countdownMessage={countdownMessage} ... ./>
```

packages/blockchain-wallet-v4/src/network/api/bitpay/index.js
```diff
@@ -0,0 +1,81 @@
...
+        if (error) {
+          throw new Error(error)
+        }
+      })
+
+  /**
+   * Given a raw payment protocol body, parses it and validates it against the digest header
+   * @param rawPaymentRequest {object} Raw JSON object retrieved from the payment protocol server
+   */
+  const parsePaymentRequest = rawPaymentRequest => {
```
plondon: This looks like a lot of logic to be doing inside of a request

packages/blockchain-wallet-v4-frontend/src/data/goals/sagas.js
```diff
@@ -127,6 +136,57 @@ export default ({ api }) => {
...
+    yield put(actions.goals.deleteGoal(id))
+
+    yield call(getBtcBalance)
+    const { r } = data
+    const invoiceId = r.split('/i/')[1]
+    const currency = yield select(selectors.core.settings.getCurrency)
+    const btcRates = yield select(selectors.core.data.btc.getRates)
+
+    const rawPaymentRequest = yield call(api.getRawPaymentRequest, invoiceId)
+    const paymentRequest = yield call(
```
plondon: What happens if this request fails?

packages/blockchain-wallet-v4-frontend/src/modals/Btc/SendBtc/FirstStep/index.js
```diff
@@ -15,7 +15,6 @@ class FirstStep extends React.Component {
 
   render () {
     const { data, actions } = this.props
-
     return data.cata({
       Success: value => (
         <Success
```
plondon: Maybe a paypro ? PayProSuccess : Success template would be better, instead of all the logic in template.success.js.

plondon:
Just tried with bitcoin:?r=https://test.bitpay.com/i/Jr629pwsXKdTCneLyZja4t and because`This invoice was not found or has been archived` endpoint was 404. We should display some error, more reason to move that api logic to a saga.

packages/blockchain-wallet-v4-frontend/src/components/Form/CountdownTimer/template.js
will-blockchain: Reason it was done like this might be the way this formatted message element works, I'm getting the error that 
Messages must be statically evaluate-able for extraction.
  36 |         <FormattedMessage
  37 |           id='modals.sendbtc.firststep.paypro_expiry_text'
> 38 |           defaultMessage={countdownMessage + timeLeft}

Not sure how to do this honestly, can take another stab at it later

packages/blockchain-wallet-v4-frontend/src/modals/Btc/SendBtc/FirstStep/index.js
will-blockchain: I was thinking that too, guessing you mean like a new PayProSuccess template? I can do that but was thinking I'd be repeating a lot of code that's already in template.success.js (Success) if I'm understanding you correctly, figured it might be better to just add a couple tidbits of paypro specific stuff inside the existing template instead of creating a whole new first step template with much of the same stuff thats in Success

packages/blockchain-wallet-v4/src/network/api/bitpay/index.js
will-blockchain: switched this function to a new bitpay saga

packages/blockchain-wallet-v4-frontend/src/data/goals/sagas.js
will-blockchain: added catch block

will-blockchain:
> Just tried with bitcoin:?r=https://test.bitpay.com/i/Jr629pwsXKdTCneLyZja4t and because`This invoice was not found or has been archived` endpoint was 404. We should display some error, more reason to move that api logic to a saga.

So the flow for the web wallet integration is someone will have a bitpay invoice open and click the Pay in Wallet link or paste the bitcoin:?r=... url in their browser, and they only way they will be able to get that link or click that Pay in Wallet button is if the invoice is still active, as soon as the 15 minute timer is up the BitPay invoice UI changes and you can no longer click the button or copy the link, feel like it's a very rare case that someone would ever be able to click/go to a bitcoin:?r=... link with an expired invoice as BitPay's UI doesn't allow that at all, so just logging errors to the action.logs.logErrorMessage might be fine. If you disagree, I can spin up some sort of error modal maybe

packages/blockchain-wallet-v4-frontend/src/components/Form/CountdownTimer/template.js
plondon: Yes you're right about FormattedMessage but countdownMessage would be a react component (`Text`).

packages/blockchain-wallet-v4-frontend/src/modals/Btc/SendBtc/FirstStep/index.js
plondon: Yeah, the more I think about it the more I think I like you're approach more.

plondon:
> If you disagree, I can spin up some sort of error modal maybe

Yeah we should handle the error, can be a simple alert action.

packages/blockchain-wallet-v4-frontend/src/data/components/sendBtc/sagas.js
```diff
@@ -66,10 +67,9 @@ export default ({ coreSagas, networks }) => {
         payment = yield payment.from(defaultIndex, ADDRESS_TYPES.ACCOUNT)
         if (to) payment = yield payment.to(to)
       }
-      const defaultFeePerByte = path(
-        ['fees', feeType || 'regular'],
-        payment.value()
-      )
+      const defaultFeePerByte = payPro
+        ? (payPro.feePerByte + 1) | 0
```
schnogz: Not sure I'm following this logic. in what situation would this bitwise ever result in the "or" condition (i.e 0)? 

(`null`, `undefined`) + 1 will be 1 and therefor the bitwise seems unnecessary.  is this intended?

packages/blockchain-wallet-v4-frontend/src/data/goals/sagas.js
```diff
@@ -85,12 +86,21 @@ export default ({ api }) => {
...
     const decodedPayload = decodeURIComponent(pathname + search)
     const bip21Payload = bip21.decode(decodedPayload)
-    const { address } = bip21Payload
-    const { amount, message } = bip21Payload.options || {}
-    const data = { address, amount, description: message }
-    yield put(actions.goals.saveGoal('payment', data))
-    yield put(actions.router.push('/wallet'))
-    yield put(actions.alerts.displayInfo(C.PLEASE_LOGIN))
+    // check for BitPay payment protocol
+    if (bip21Payload.options && bip21Payload.options.r) {
```
schnogz: prefer Ramda => `if(path(['options', 'r'], bip21Payload)) {`

packages/blockchain-wallet-v4-frontend/src/data/goals/sagas.js
```diff
@@ -85,12 +86,21 @@ export default ({ api }) => {
...
+      const { r } = bip21Payload.options || {}
+      const data = { r }
```
schnogz: confusing intent here. need to make these two lines be more succinct

here is a start => `const r = pathOr({}, ['options', 'r'], bip21Payload)`.  but what is line 92 doing?

packages/blockchain-wallet-v4-frontend/src/modals/Btc/SendBtc/FirstStep/template.success.js
```diff
@@ -235,90 +272,126 @@ const FirstStep = props => {
...
+                ) : (
+                  <FormattedMessage
+                    id='modals.sendbtc.firststep.customizefee'
+                    defaultMessage='Customize Fee'
+                  />
+                )}
+              </Link>
+            </ColRight>
+          </FeeFormGroup>
+          {feePerByteToggled ? (
```
schnogz: `{feePerByteToggled && (` then get rid of line 375

packages/blockchain-wallet-v4/src/network/api/bitpay/index.js
```diff
@@ -0,0 +1,36 @@
+import axios from 'axios'
+
+const url = 'https://bitpay.com/'
```
schnogz: i'd prefer this to be in wallet options `domains` and then passed in. @plondon thoughts?

packages/blockchain-wallet-v4-frontend/src/data/components/sendBtc/sagas.js
will-blockchain: Logic is, if new saga has payPro information, then set defaultFeePerByte to (paypro + 1) | 0, (the |0 truncates the decimal places (clean way to do it),  so an int value can be passed in as feePerByte), if new send saga does not have payPro then go forward with regular feePerByte setting

packages/blockchain-wallet-v4-frontend/src/modals/Btc/SendBtc/FirstStep/template.success.js
will-blockchain: I didn't write that code, didnt want to touch something that was already there

packages/blockchain-wallet-v4-frontend/src/data/goals/sagas.js
will-blockchain: I followed the existing code in the current defineSendBtcGoal to create r
`const { amount, message } = bip21Payload.options || {}
and line 92 creates a new data object with r as a field to be passed into the saveGoal call

packages/blockchain-wallet-v4-frontend/src/components/Form/CountdownTimer/template.js
will-blockchain: Looking at doing this now, but I've also got the countdown timer on the second step of the send flow though so I'd have to create two new countdownMessage react components in each respective step and find a way to handle the getCountdownColor without including that function in both steps as well. Since both steps use the same countdown message feel like its better to keep it as a singular thing in the countdowntimer template.

packages/blockchain-wallet-v4-frontend/src/data/goals/sagas.js
will-blockchain: addressed in newest commit

packages/blockchain-wallet-v4-frontend/src/data/goals/sagas.js
will-blockchain: creates a new data object with the bitpay link r as the field, to be passed into the runPaymentProtocol goal saga

will-blockchain: add pathOr in latest

will-blockchain:
> > If you disagree, I can spin up some sort of error modal maybe
> 
> Yeah we should handle the error, can be a simple alert action.

Cool no problem added an alert in the latest commit.

will-blockchain:
Also need to fix the amount/description bug currently in production before this will work. If you go through a payment protocol link like bitcoin:?address=1KLENCHHEJDLL87DLJ6...?amount=1 and then after the send modal pops up populating your info and you click continue to go to the second step, amount is undefined and isn't getting passed properly. (description isnt either if it isn't edited)

packages/blockchain-wallet-v4/src/network/api/bitpay/index.js
plondon: Yes needs to be in wallet-options. Please create PR here https://github.com/blockchain/wallet-options

plondon: You just need to change wallet-options-v4.json for dev/staging/prod/testnet

will-blockchain: switched to wallet options domains in latest commit

will-blockchain:
Fixed bug with amount not being passed in to second step that currently affects production, after discussion with Mark changed fee rate from custom fee per byte [bitpay based] to priority, and addressed @schnogz wallet options domain comment. Everything ready to go 🚀 @plondon

PR #1773, opened 2019-08-19
Feature/alert toast animation

jjBlockchain:
Moved props and color variable above useEffect in Toast.js


jjBlockchain:
@schnogz and @lzhuor I just moved the props and color variable above useEffect, everything else looks good to me. Can you guys please review to make sure?

PR #1779, opened 2019-08-20
Feat(Segwit)  🎉

plondon:
## Description (optional)
Update payload to v4 for ability to have multiple derivation types

## Testing Steps (optional)

Fixes WEB-130
Fixes WEB-131


packages/blockchain-wallet-v4-frontend/src/data/components/manageAddresses/actions.js
```diff
@@ -1,92 +1,122 @@
 import * as AT from './actionTypes'
 
 // UNUSED ADDRESSES
-export const deleteAddressLabel = (accountIdx, walletIdx, addressIdx) => ({
+export const deleteAddressLabel = (
```
plondon: Just wondering if we need to pass derivation to all these functions @schnogz or can we figure that out deeper down in the core?

plondon: Nevermind, we need to be explicit here, you're right

packages/blockchain-wallet-v4/src/coinSelection/coin.js
```diff
@@ -48,6 +30,20 @@ export class Coin extends Type {
...
+          return 'P2PKH'
+        case '3':
+          return 'P2SH-P2WPKH'
+        default:
+          return 'P2PKH'
+      }
+    } catch (e) {
+      return 'P2PKH'
+    }
+  }
```
plondon: Is the best way to do this?

plondon:
Fixes WEB-130

packages/blockchain-wallet-v4/src/walletCrypto/index.spec.js
```diff
@@ -148,7 +151,9 @@ describe('WalletCrypto', () => {
     })
     it('should fail because of wrong password', done => {
       wCrypto.decryptWallet('wrong password', data.v3).fork(failure => {
-        expect(failure).toEqual('v2v3: wrong_wallet_password')
+        expect(failure.message).toEqual(
+          'Unsupported state or unable to authenticate data'
+        )
         done()
       }, done)
     })
```
akly: Looks like these are decryption unit tests? Can I suggest two more test cases:

1. Corrupted ciphertext - take a correctly encrypted payload and tag, flip a single bit in the payload
2. Corrupted GCM tag - take a correctly encrypted payload and tag, flip a single bit in the tag

Expected result - should return exactly the same error as incorrect password

packages/blockchain-wallet-v4/src/coinSelection/coin.js
sidhujag: What about native segwit? Is native segwit not going to be supported in this PR?

packages/blockchain-wallet-v4/src/types/HDAccount.js
```diff
@@ -1,27 +1,41 @@
...
-import * as crypto from '../walletCrypto'
-import { fromJS as iFromJS } from 'immutable-ext'
+import { fromJS as iFromJS } from 'immutable-ext' // if we delete this import, wallet tests will fail -  ¯\_(ツ)_/¯
+/* eslint-disable */
+import { over, traversed, traverseOf, view } from 'ramda-lens'
+import Task from 'data.task'
 import Type from './Type'
-/* eslint-enable */
+
+export const DEFAULT_DERIVATION_TYPE = 'segwitP2SH'
```
sidhujag: bitcoin core now uses native segwit as default, this should be the same I think... default should probably be p2wpkh / p2wsh

plondon: Thank you for having a look at the PR. Native segwit will come after this PR.

The majority of the work in this PR is to support multiple derivation types per account. Once we have segwitP2SH support adding native segwit becomes much easier. We could support native segwit as the default deriviation type one day, but there are still some services that don't support sending to native segwit, that is why it should be opt in for now.

sidhujag: do you mean spending from native segwit (bech32)? which services are linked to the wallet that don't support? I guess bip84 would be then used instead of 44 in that PR.. maybe if those services don't support I can take a look and perhaps get in touch to see when they will support or help out if its open source.

plondon: I mean sending to native segwit, some services don't support that. Only giving users a native segwit address to receive from would lock them into only using services that support sending to native segwit. Binance, for example, does not support sending to native segwit.

sidhujag: Ah I see so you have to use legacy by default for now. Got it.

sidhujag: also one more thing, if your supporting bip49 then will api work with ypub?

sidhujag: Maybe descriptors should be used instead to avoid xpub ypub zpub hell: https://github.com/bitcoin/bitcoin/blob/master/doc/descriptors.md

plondon: The API (multiaddr and unspent) will accept an optional parameter activeP2SH for the xpubs derived from the segwitP2SH chain (49)

sidhujag: Cool maybe it can be like ?bip49=xpub3|xpub4 so when bip84 comes it will be another extension.

schnogz:
@plondon we might want to deal with this merge from `development` branch before we start picking up SW again.

Afsaneh5054 (approved):

Afsaneh5054 (approved):

Afsaneh5054 (approved):

Afsaneh5054 (approved):

packages/blockchain-wallet-v4/src/coinSelection/coin.js
schnogz: yes we are upgrading and defaulting to native for this release

plondon:
Notes on adding v3=>v4 upgrade process:

1. don't change payload
2. leave `derivations` params, but ignore them?

packages/blockchain-wallet-v4-frontend/src/data/auth/sagas.js
```diff
@@ -111,11 +123,22 @@ export default ({ api, coreSagas }) => {
...
       // If needed, the user should upgrade its wallet before being able to open the wallet
       const isHdWallet = yield select(selectors.core.wallet.isHdWallet)
       if (!isHdWallet) {
-        yield call(upgradeWalletSaga)
+        yield call(upgradeWalletSaga, isDoubleEncrypted, 3)
       }
+      const isLatestVersion = yield select(
+        selectors.core.wallet.isWrapperLatestVersion
+      )
+      if (!isLatestVersion) {
```
plondon, line 137: && flag

packages/blockchain-wallet-v4/src/types/HDAccount.ts
```diff
@@ -34,72 +47,147 @@ export class HDAccount extends Type {}
...
+  if (is(HDAccount, account)) {
+    return account
+  }
+
+  const accountCons = compose(
+    over(derivations, DerivationList.fromJS),
+    // @ts-ignore
+    a => new HDAccount(a),
+    assoc('index', index),
+    migrateFromV3
```
plondon, line 183: flag check

packages/blockchain-wallet-v4/src/types/HDAccountList.js
```diff
@@ -21,8 +31,19 @@ export const selectByXpub = curry((xpub, as) =>
...
+  )
+  return map(
+    pluck('xpub'),
+    groupBy(c => c.type, flatten(activeAccounts.toJS()))
+  )
+})
+
 export const selectContext = pipe(HDAccountList.guard, accList => {
-  return map(HDAccount.selectXpub, filter(HDAccount.isActive, accList))
+  return map(HDAccount.selectAllXpubs, filter(HDAccount.isActive, accList))
```
plondon, line 46: flag check

packages/blockchain-wallet-v4/src/coinSelection/coinSelection.spec.js
```diff
@@ -5,13 +5,171 @@ import * as cs from './index'
...
+      it('should return the right transaction size (1 P2PKH, 1 P2PKH)', () => {
+        // 10 + 148 + 34 = 192
+        expect(cs.transactionBytes([p2pkhIn], [p2pkhOut])).toEqual(192)
+      })
+      it('should return the right transaction size (1 P2PKH, 1 P2WPKH)', () => {
+        // 10 + 148 + 31 = 189
+        expect(cs.transactionBytes([p2pkhIn], [p2wpkhOut])).toEqual(189)
+      })
+      it('should return the right transaction size (1 P2WPKH, 1 P2PKH)', () => {
+        // 10.75 + 67.75 + 34 = 189
```
pavel-bc, line 30:
```suggestion
        // 10.75 + 67.75 + 34 = 112.5
```

TheLeoB (commented):
will this only be included with the segwit release then?

packages/blockchain-wallet-v4-frontend/package.json
```diff
@@ -87,7 +87,7 @@
     "bip21": "2.0.2",
     "bip39": "2.5.0",
     "bitcoinforksjs-lib": "git://github.com/blockchain/bitcoinjs-lib.git#opt-in-bitcoincash-sighash",
-    "bitcoinjs-lib": "3.1.1",
+    "bitcoinjs-lib": "5.1.6",
```
milan-bc, line 90: should we use latest `5.1.10` or `5.2`?

schnogz (approved):

amadeo-bc:
🔥🔥🔥🔥🔥

PR #1795, opened 2019-08-30
Feat/intro tour

jjBlockchain:
## Description (optional)
Created an introduction tour for new users.


schnogz:
Overall really nice work! 🎉 

Please also rename assets and code instances from 'welcome' to walletTour'.  That includes the goal name, modal folder and file names, etc.

PR #1810, opened 2019-09-06
Airdrop message updated

jjBlockchain:
When a user is under review, added an updated notice for airdrops, also changed out the alert-icon with a circle filled in an alert icon.


Adam16p:
Duplicate of #

PR #1835, opened 2019-09-13
Fix XLM Socket

seanblockchain:
## Description 
1. Moved `actions.middleware.webSocket.xlm.startStreams()` to execute before `coreSagas.kvStore.xlm.fetchMetadataXlm` in order to avoid race condition with setting `onMessage` and `onError` attributes of the `HorizonStreamingService`.
2. Added test for `startStreams()`.

## Testing Steps
1. Open up two wallet sessions in different tabs. 
2. Send xlm from once account to the other.
3. Check to make sure alert popup fires, and balance updates without refreshing.

### DIFF
--- a/packages/blockchain-wallet-v4-frontend/src/data/auth/sagas.js
+++ b/packages/blockchain-wallet-v4-frontend/src/data/auth/sagas.js
@@ -93,7 +93,6 @@ export default ({ api, coreSagas }) => {
     yield put(actions.middleware.webSocket.bch.startSocket())
     yield put(actions.middleware.webSocket.btc.startSocket())
     yield put(actions.middleware.webSocket.eth.startSocket())
-    yield put(actions.middleware.webSocket.xlm.startStreams())
   }
 
   const authNabu = function * () {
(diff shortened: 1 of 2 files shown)


packages/blockchain-wallet-v4-frontend/src/data/auth/sagas.js
```diff
@@ -123,6 +122,7 @@ export default ({ api, coreSagas }) => {
         coreSagas.kvStore.eth.fetchMetadataEth,
         askSecondPasswordEnhancer
       )
+      yield put(actions.middleware.webSocket.xlm.startStreams())
```
plondon: Does the stream need any xlm data from metadata (the next call), such as the user's account?

seanblockchain: @plondon `startStreams` does not need any account info. It simply hits the `open` fn in `HorizonStreamingService` which defines `onMessage` and `onError`.

`_subscribeToAccount` is fired from the following metadata call which initiates the stream with the fetched accountId.

PR #1837, opened 2019-09-13
Fix: signup clean

seanblockchain:
## Description 
1. Added checkmark svg to alerts with `success` status.
2. Added opacity on-hover highlighting to "Don't have a wallet?".

## Testing Steps (optional)
1. Go to signup, mouse over "Don't have a wallet?". 
2. Submit send, notice alert popup with checkmark.


packages/blockchain-wallet-v4-frontend/src/components/Alerts/template.js
```diff
@@ -51,6 +52,23 @@ const Alerts = props => {
...
+      xmlns='http://www.w3.org/2000/svg'
+    >
+      <path
+        d='M21.3333 0H2.66667C1.18667 0 0 1.2 0 2.66667V21.3333C0 22.8 1.18667 24 2.66667 24H21.3333C22.8133 24 24 22.8 24 21.3333V2.66667C24 1.2 22.8133 0 21.3333 0ZM9.33333 18.6667L2.66667 12L4.54667 10.12L9.33333 14.8933L19.4533 4.77333L21.3333 6.66667L9.33333 18.6667Z'
+        fill='#0C6CF2'
+      />
+    </svg>
+  )
+}
+
```
plondon: Should we add an icon?

schnogz: yeah, let's add it as an icon just to be safe and keep things clean.  @seanblockchain let me know if you need help doing this and i can walk you through the "process" ;)

packages/blockchain-wallet-v4-frontend/src/components/Alerts/template.js
```diff
@@ -43,6 +43,13 @@ const Alerts = props => {
             persist={persist}
             onClose={() => handleClose(id)}
           >
+            {nature === 'success' ? (
```
schnogz: By adding this check here, now every single alert that is of nature `success` gets this checkmark icon.  Is this really what design wanted?

For example, now on language change gets the icon.
Screen Shot 2019-09-17 at 9 45 21 AM

Also the green edge and blue icon seems strange... 🤔

seanblockchain: Yes, agreed wrt blue on green. 

Also, I assume the checkmark is actually only meant to appear on signup.

I could add it directly to the alert component (`src/components/Alerts/messages.js`) FormattedMessage for that case.

PR #1852, opened 2019-09-20
fix(webpack-dev-server): Fix CSP nonce error in console.

Empowerful:
Replace `**CSP_NONCE**` in index.html using the webpack-dev-server (as is done in production).

## Description (optional)

Finally fix the error message that always comes up first in the console when running locally:

`Refused to execute inline script because it violates the following Content Security Policy directive: "script-src 'self' 'unsafe-eval'". Either the 'unsafe-inline' keyword, a hash ('sha256-yZmdgVvzROiMMhereftUxrofbQsrGpeQmdlyusoGIp4='), or a nonce ('nonce-...') is required to enable inline execution.`

## Testing Steps (optional)

Start the application locally and open the JavaScript console.  Notice the absence of the above error.


packages/blockchain-wallet-v4-frontend/package.json
```diff
@@ -145,5 +145,8 @@
     "slick-carousel": "1.8.1",
     "styled-components": "4.2.0",
     "zxcvbn": "4.4.2"
+  },
+  "devDependencies": {
+    "html-webpack-string-replace-plugin": "1.1.0"
```
schnogz: dev deps go in the root level `package.json`

PR #1853, opened 2019-09-20
feat(price-chart-cta)

jjBlockchain:
Web - 2716


packages/blockchain-wallet-v4-frontend/src/scenes/Home/PriceChart/Footer/index.js
```diff
@@ -0,0 +1,95 @@
...
+      case 'BTC':
+        return setCoinName('Bitcoin')
+      case 'ETH':
+        return setCoinName('Ethereum')
+      case 'BCH':
+        return setCoinName('Bitcoin Cash')
+      case 'XLM':
+        return setCoinName('Stellar')
+      default:
+        setCoinName('Bitcoin')
```
plondon: Should not hardcode coin names, they are all set in wallet-options. Pass wallet options into this component and use that instead.

packages/blockchain-wallet-v4-frontend/src/scenes/Home/PriceChart/Footer/index.js
```diff
@@ -0,0 +1,95 @@
...
+    }
+  }, [priceChart.coin])
+
+  return (
+    <Wrapper>
+      {coinifySupported && (
```
plondon: Only BTC is supported currently for Buy and Sell

PR #1884, opened 2019-10-10
feat/(price-chart-cta): added price chart button cta

jjBlockchain:
web# 2716

### DIFF
--- a/packages/blockchain-wallet-v4-frontend/src/data/analytics/model.js
+++ b/packages/blockchain-wallet-v4-frontend/src/data/analytics/model.js
@@ -101,8 +101,3 @@ export const PIT_EVENTS = {
   CONNECT_NOW: ['pit', 'link_page', 'connect_now_click'],
   LEARN_MORE: ['pit', 'link_page', 'learn_more_click']
 }
-
-export const PRICE_CHART_EVENTS = {
-  CLICK_BUY_BITCOIN: ['price_chart', 'footer_button', 'buy_bitcoin'],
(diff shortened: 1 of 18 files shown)


packages/blockchain-wallet-v4-frontend/src/data/analytics/model.js
```diff
@@ -101,3 +101,8 @@ export const PIT_EVENTS = {
   CONNECT_NOW: ['pit', 'link_page', 'connect_now_click'],
   LEARN_MORE: ['pit', 'link_page', 'learn_more_click']
 }
+
+export const PRICE_CHART_EVENTS = {
+  CLICK_BUY_BITCOIN: ['price_chart', 'footer_button', 'buy_bitcoin'],
+  CLICK_SWAP_COIN: ['price_chart', 'footer_button', 'swap_coin']
+}
```
plondon: These are unnecessary since matomo refactor

plondon:
Looks better and it's nice that there is less logic.

However, designs have chart extending to the bottom of the container.

You also need to consider what happens if both !silver and user can't buy.

Screen Shot 2019-10-14 at 9 19 13 AM

jjBlockchain:
> Looks better and it's nice that there is less logic.
> 
> However, designs have chart extending to the bottom of the container.
> 
> You also need to consider what happens if both !silver and user can't buy.
> 
> Screen Shot 2019-10-14 at 9 19 13 AM

I will confirm with the design team on this

PR #1890, opened 2019-10-15
feat(pit-linked-info-icon): added info icon when user's wallet is lin…

jjBlockchain:
…ked to pit
WEB-125


packages/blockchain-wallet-v4-frontend/src/layouts/Wallet/MenuLeft/Navigation/template.js
```diff
@@ -270,7 +278,7 @@ const Navigation = props => {
             target='_blank'
             style={{ width: '100%' }}
           >
-            {renderPitSidenav()}
+            {renderPitSidenav(null, true)}
```
schnogz: can you just pass `props.isPitAccountLinked` instead of `true`?  it will be slightly more readable

PR #1891, opened 2019-10-15
Fix/segwit legacy toggle

jjBlockchain:
Web-118

### DIFF
--- a/packages/blockchain-wallet-v4-frontend/src/scenes/Settings/Addresses/Btc/ManageAddresses/UnusedAddresses/index.js
+++ b/packages/blockchain-wallet-v4-frontend/src/scenes/Settings/Addresses/Btc/ManageAddresses/UnusedAddresses/index.js
@@ -32,6 +32,14 @@ const ClickableText = styled(Text)`
 const DerivationSwitchContainer = styled.div`
   margin-left: 12px;
 `
+const LegacyToggledLink = styled(Link)`
+  color: ${({ derivation, theme }) =>
+    derivation === 'legacy' ? theme.grey000 : theme['brand-secondary']};
+`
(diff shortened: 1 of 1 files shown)


packages/blockchain-wallet-v4-frontend/src/scenes/Settings/Addresses/Btc/ManageAddresses/UnusedAddresses/index.js
```diff
@@ -32,6 +32,14 @@ const ClickableText = styled(Text)`
...
   margin-left: 12px;
 `
+const LegacyToggledLink = styled(Link)`
+  color: ${({ derivation, theme }) =>
+    derivation === 'legacy' ? theme.grey000 : theme['brand-secondary']};
+`
+const SegWitToggledLink = styled(Link)`
+  color: ${({ derivation, theme }) =>
+    derivation === 'segwitP2SH' ? theme.grey000 : theme['brand-secondary']};
+`
```
schnogz: extremely redundant code between the two styled components. what about something like the following?

```
const ToggledLink = styled(Link)`
  color: ${({ isToggled, theme }) =>
    isToggled ? theme.grey000 : theme['brand-secondary']};
```

then in the legacy component (and similar for SW component)
```
<ToggledLink
  isToggled={derivation === 'legacy'}
  weight={500}
  size='13px'
  data-e2e='btcManageLegacyWalletLink'>
       ...
</ToggledLink>
```

plondon:
There is an even simpler way to target the element whose color you want to change:

Screen Shot 2019-10-16 at 10 53 23 AM

Use the `active` class set by @blockchain/components

PR #1896, opened 2019-10-17
fix(authorize-login)

jjBlockchain:
WEB-11


packages/blockchain-info-components/src/Buttons/Button.js
```diff
@@ -106,6 +106,12 @@ const selectColor = (nature, disabled, small) => {
         backgroundColor: 'brand-secondary',
         borderColor: 'brand-secondary'
       }
+    case 'authorize-login-approve':
```
schnogz: 1) How about naming it just `green` or something more generic?
2) Always prefer camel casing unless working in `kvStore` or other legacy APIs

schnogz:
Looks great overall!  Did UX team sign off on this "new design" due to the minimal dataset?

packages/blockchain-info-components/src/Buttons/Button.js
qiornu: замените прокладку между компьютером и креслом, кракозябры кругом.

PR #1898, opened 2019-10-17
fix(duplicate-wallet-labels)

jjBlockchain:
Web4-2145

### DIFF
--- a/packages/blockchain-wallet-v4-frontend/src/data/modules/addressesBch/sagas.js
+++ b/packages/blockchain-wallet-v4-frontend/src/data/modules/addressesBch/sagas.js
@@ -1,6 +1,7 @@
 import { select, call, put } from 'redux-saga/effects'
 import { actions, selectors } from 'data'
 import * as C from 'services/AlertService'
+import { requireUniqueWalletName } from 'services/FormHelper'
 import { promptForInput } from 'services/SagaService'
 import { utils } from 'blockchain-wallet-v4/src'
 const { toCashAddr } = utils.bch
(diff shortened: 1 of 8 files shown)


packages/blockchain-wallet-v4-frontend/src/data/modules/addressesBch/sagas.js
```diff
@@ -12,10 +12,22 @@ export default ({ coreSagas, networks }) => {
   const editBchAccountLabel = function * (action) {
     try {
       const { index, label } = action.payload
+      const wallets = (yield select(
+        selectors.core.common.bch.getAccountsBalances
```
schnogz: I think the more appropriate selector would be `selectors.core.common.bch.getActiveHDAccounts`

packages/blockchain-wallet-v4-frontend/src/data/wallet/sagas.js
```diff
@@ -52,10 +53,20 @@ export default ({ coreSagas }) => {
...
       let { index, label } = action.payload
+      const wallets = (yield select(selectors.core.common.btc.getHDAccounts))
+        .getOrFail()
+        .map(wallet => wallet.label)
+      const isUnique = (value, allValues) => {
+        const walletIdx = wallets.indexOf(value)
+        return walletIdx !== index && walletIdx > -1
+          ? 'Wallet name is already taken.'
+          : undefined
+      }
```
schnogz: to reduce redundancy, lets rename to be `requireUniqueWalletName` and then export it from
`packages/blockchain-wallet-v4-frontend/src/services/FormHelper/validators.js` and import into both sagas.

PR #1902, opened 2019-10-21
RFC: Feature TypeScript

plondon:
cc @schnogz @Lyncee59 @blockolly 

## Description (optional)
This is a RFC and starting point for a TypeScript integration into the wallet. So far this PR is doing very little as I am only trying to set the groundwork.

Here is how I am trying to breakdown the work.
1. Create Typed (`.d.ts`) files of our API requests and responses.
2. Consume API responses in `sagas.ts` files
3. Define data types for selectors
4. Consume data from selectors in index.js and template.js files.

So far I have only touched on items 1 and 2.

### 1. Create Typed (`.d.ts`) files of our API requests and responses.
`data/btc/sagas.ts` requests data from 3 different `network/api`'s [network/api/btc, network/api/wallet, network/api/misc].

Because api is passed around as an object (due to wallet-options requirement before app start) `data/btc/sagas.ts` cannot know about the different api types, therefore we have to manually generate these files (for now). See [network/api/btc/BtcAPI.d.ts](https://github.com/blockchain/blockchain-wallet-v4-frontend/compare/feat/ts?expand=1#diff-4760a9f359cbb22ed08d2f6971e37c4c), [network/api/misc/MiscAPI.d.ts](https://github.com/blockchain/blockchain-wallet-v4-frontend/compare/feat/ts?expand=1#diff-8e6858d6c30a18fd5c5dea6b56b3365c) and [network/api/WalletAPI.d.ts](https://github.com/blockchain/blockchain-wallet-v4-frontend/compare/feat/ts?expand=1#diff-529994ed46c788ab71616a5288126c90).

These files were generated using the command:

`tsc packages/blockchain-wallet-v4/src/network/api/btc/index.ts --declaration --emitDeclarationOnly`, I then changed the generated file name to BtcAPI.d.ts and changed the export default to export type. I would like to automate this process, see remaining work.

### 2. Consume API responses in `sagas.ts` files
Currently, the only TypeScript-ish thing we are using right now is the [api.fetchBlockchainData](https://github.com/blockchain/blockchain-wallet-v4-frontend/compare/feat/ts?expand=1#diff-529994ed46c788ab71616a5288126c90R6-R17) call. We export API from `network/types.ts` and now api.fetchBlockchainData is type-safe. However, any change to network/api/wallet.ts and we will need to redo step 1, which is not ideal. 

Remaining work and questions:
- [ ] Add information and instructions to bwv-frontend/help
- [x] Automate `.d.ts` file generation for API types.
- [ ] Generate type file for Remote types.
- [ ] Add https://github.com/codechecks/typecov
- [x] What to do about reducers?
- [ ] What to do about sagas.spec.js tests?
- [ ] Make sure yarn ci:compile works


package.json
```diff
@@ -115,6 +121,11 @@
     "@storybook/cli": "4.1.9",
     "@storybook/react": "4.1.9",
     "@storybook/storybook-deployer": "2.8.1",
+    "@types/jest": "24.0.19",
+    "@types/node": "12.11.1",
+    "@typescript-eslint/eslint-plugin": "2.4.0",
+    "@typescript-eslint/parser": "2.4.0",
+    "awesome-typescript-loader": "5.2.1",
```
Lyncee59: I saw different typescript loaders, this one and at least another one, are we sure its the best loader to use ? (is it well maintained or deprecated?)

packages/blockchain-wallet-v4-frontend/package.json
```diff
@@ -71,7 +71,8 @@
       "<rootDir>/../../node_modules/enzyme-to-json/serializer"
     ],
     "transform": {
-      "^.+\\.(js|jsx)$": "babel-jest"
+      "^.+\\.(js|jsx)$": "babel-jest",
+      "^.+\\.tsx?$": "ts-jest"
```
Lyncee59: We should also add the .ts extension here, typically files which don't contain any JSX should be .ts files, and not .tsx

packages/blockchain-wallet-v4/src/network/api/btc/BtcAPI.d.ts
```diff
@@ -0,0 +1,10 @@
+export type BtcAPI = {
+  getBtcTicker: () => any
```
Lyncee59: Overall, the main benefits here would be to produce the data models interfaces that we expect coming from our api endpoints to get full benefits of TS. I know its just a small POC, but to think about it for later !

packages/blockchain-wallet-v4/src/redux/data/btc/types.ts
```diff
@@ -0,0 +1,6 @@
+export interface FetchTransactionsAction {
```
Lyncee59: I would keep this type in the sagas file and export it, then join all the types coming from actions.ts, actionTypes.ts, reducers.ts etc... in this top level types.ts file.

package.json
plondon: Yeah, both ts-loader and awesome-typescript-loader are well maintained and used by thousands of people. awesome-typescript loader was faster for me so I went with that.

packages/blockchain-wallet-v4-frontend/package.json
plondon: It should handle both 👍

packages/blockchain-wallet-v4/src/network/api/btc/BtcAPI.d.ts
plondon: Yes, 100% agree, the goal would be to generate these automatically and do the actual work in the respective index.ts files.

packages/blockchain-wallet-v4/src/redux/data/btc/types.ts
plondon: Agree 100%

plondon:
In the initial plan I was thinking of creating .d.ts files for apis. I think actually a better plan would be to create the APIs inside the sagas, instead of passing them in. In the case of data/btc for example:

1. We add TypeScript to api/x/index.ts files and export default (nothing changes except adding TS)
2. In sagas we import these files directly like:

sagas.ts
```
import BTCApi from '../../../network/api/btc/index'
export default rest => {
  const api = BTCApi({ ...rest })
```

However, **first we need to clean up network/api and create wallet api**. Next PRs will be doing that.

package.json
```diff
@@ -115,6 +121,13 @@
     "@storybook/cli": "4.1.9",
     "@storybook/react": "4.1.9",
     "@storybook/storybook-deployer": "2.8.1",
+    "@types/jest": "24.0.19",
+    "@types/node": "12.11.1",
+    "@types/react": "^16.9.11",
+    "@types/react-dom": "^16.9.3",
+    "@typescript-eslint/eslint-plugin": "2.4.0",
+    "@typescript-eslint/parser": "2.4.0",
+    "awesome-typescript-loader": "5.2.1",
```
plondon: Pin deps

plondon:
Closing for now https://linear.app/issue/WEB-96

lzhuor:
🔥 Nice one.

PR #1934, opened 2019-11-04
Debt/bitcoinjs

seanblockchain:
## Upgrade to bitcoinjs-lib `5.1.6`
Upgrades to core and frontend in accordance with the library changes.

## Testing Steps
1. Create new wallet. Check that all elements of the kv store are successful kvStore->*. Check that the wallet has been created in wallet walletPath->wallet->hd_wallets[0].
2. Import legacy address. Check walletPath->wallet->addresses.
3. Add new wallet. Check walletPath->wallet->hd_wallets[1].
4. Connect lockbox. 
5. Sends for btc, bch, stellar, eth, and paxos.
6. Sign message in wallets/addresses settings.
7. Show private key in wallets/addresses settings.


packages/blockchain-wallet-v4/src/types/KVStoreEntry.js
```diff
@@ -83,16 +81,16 @@ export const deriveMetadataNode = masterHDNode => {
...
-)
+export const fromMetadataXpriv = curry((xpriv, typeId, network) => {
+  return fromMetadataHDNode(Bitcoin.bip32.fromBase58(xpriv, network), typeId)
+})
 
 export const fromMetadataHDNode = curry((metadataHDNode, typeId) => {
   let payloadTypeNode = metadataHDNode.deriveHardened(typeId)
-  let node = payloadTypeNode.deriveHardened(0)
-  let privateKeyBuffer = payloadTypeNode.deriveHardened(1).keyPair.d.toBuffer()
+  let privateKeyBuffer = payloadTypeNode.deriveHardened(1).__D
```
mpfluger: This does not look equivalent to what happened in the old one.

packages/blockchain-wallet-v4/src/types/Wallet.js
```diff
@@ -549,21 +551,30 @@ export const getHDPrivateKeyWIF = curry(
...
         )
-        .map(xp => derivePrivateKey(network, xp, chain, index).keyPair.toWIF())
+        .map(xp => {
+          const node = derivePrivateKey(
+            network,
+            xp,
+            chain,
+            index
+          ).keyPair.toWIF()
+          const keypair = Bitcoin.ECPair.fromPrivateKey(node.privateKey)
```
mpfluger: Maybe I'm not getting it, but this doesn't seem right.

packages/blockchain-wallet-v4/src/utils/btc.js
```diff
@@ -1,12 +1,18 @@
...
+  bip32
+} from 'bitcoinjs-lib'
 import { equals, head, or, propOr, compose, dropLast, last } from 'ramda'
 import { decode, fromWords } from 'bech32'
-import { fromPublicKey } from 'bip32'
 import { compile } from 'bitcoinjs-lib/src/script'
 import * as OP from 'bitcoin-ops'
 import Base58 from 'bs58'
-import BigInteger from 'bigi'
+// import BigInteger from 'bigi'
```
mpfluger: Remove instead of comment

packages/blockchain-wallet-v4/src/utils/btc.js
```diff
@@ -214,7 +215,7 @@ export const calculateBalanceSatoshi = (coins, feePerByte) => {
 }
 
 export const isKey = function (btcKey) {
-  return btcKey instanceof ECPair
+  return btcKey instanceof Object // hack since ECPair instance of is breaking
```
mpfluger: How does it break?

packages/blockchain-wallet-v4/src/utils/btc.js
```diff
@@ -271,9 +272,13 @@ export const createXpubFromChildAndParent = (path, child, parent) => {
...
-  hdnode.parentFingerprint = fingerprint(pkParent)
-  hdnode.depth = pathArray.length
-  hdnode.index = last(pathArray)
-  return hdnode.toBase58()
+  let hdnode = bip32.fromPublicKey(pkChild, Buffer.from(child.chainCode, 'hex'))
+  hdnode.__PARENT_FINGERPRINT = fingerprint(pkParent)
+  hdnode.__DEPTH = pathArray.length
+  hdnode.__INDEX = last(pathArray)
+
+  return hdnode.neutered().toBase58()
```
mpfluger: Why is the `neutered()` introduced here?

packages/blockchain-wallet-v4/src/utils/btc.js
```diff
@@ -271,9 +272,13 @@ export const createXpubFromChildAndParent = (path, child, parent) => {
   let pathArray = bippath.fromString(path).toPathArray()
   let pkChild = compressPublicKey(Buffer.from(child.publicKey, 'hex'))
   let pkParent = compressPublicKey(Buffer.from(parent.publicKey, 'hex'))
-  let hdnode = fromPublicKey(pkChild, Buffer.from(child.chainCode, 'hex'))
-  hdnode.parentFingerprint = fingerprint(pkParent)
-  hdnode.depth = pathArray.length
-  hdnode.index = last(pathArray)
-  return hdnode.toBase58()
+  let hdnode = bip32.fromPublicKey(pkChild, Buffer.from(child.chainCode, 'hex'))
+  hdnode.__PARENT_FINGERPRINT = fingerprint(pkParent)
```
mpfluger: It may not be possible to change, but accessing private members here.

packages/blockchain-wallet-v4/src/walletCrypto/importExport.js
```diff
@@ -1,10 +1,11 @@
-import Bitcoin from 'bitcoinjs-lib'
+import * as Bitcoin from 'bitcoinjs-lib'
 import scrypt from 'scryptsy'
 import BigInteger from 'bigi'
 import Base58 from 'bs58'
 import Unorm from 'unorm'
 import * as WalletCrypto from './utils'
 
+// console.log('Bitcoin', Bitcoin)
```
mpfluger: Remove

packages/blockchain-wallet-v4/src/walletCrypto/importExport.js
```diff
@@ -67,7 +68,11 @@ export const parseBIP38toECPair = function (
   var AESopts = { mode: WalletCrypto.AES.ECB, padding: WalletCrypto.NoPadding }
 
   var verifyHashAndReturn = function () {
-    var tmpkey = new Bitcoin.ECPair(decrypted, null, {
+    // var tmpkey = new Bitcoin.ECPair(decrypted, null, {
```
mpfluger: Remove

packages/blockchain-wallet-v4/src/utils/btc.js
seanblockchain: Noted on accessing private members here, the bip32 api does not expose setters.
https://github.com/bitcoinjs/bip32/blob/master/src/bip32.js

packages/blockchain-wallet-v4/src/utils/btc.js
seanblockchain: Introduced neutered() since we are not interested in the private key associated with the derived node. The derivation seems to be simply used to extract the xpub from the lockbox wallet.

This could be an unnecessary step, or perhaps I am viewing the process incorrectly.

packages/blockchain-wallet-v4/src/utils/btc.js
seanblockchain: For some reason the instanceof ECPair equality is false after upgrade. Will investigate further.

packages/blockchain-wallet-v4/src/utils/btc.js
mpfluger: Yeah, I figured, just wanted to make sure

packages/blockchain-wallet-v4/src/types/Wallet.js
seanblockchain: Yes, it looks like I have an unnecessary keypair instantiation-- will remove and revert back to original code here.

packages/blockchain-wallet-v4/src/types/KVStoreEntry.js
seanblockchain: Good call, I am not passing through the derived node at index 0 into the KV store.

Is there a reason why we are deriving another node for the encryption key at index 1 instead of just reusing the 0 indexed node?

packages/blockchain-wallet-v4/src/types/Wallet.js
seanblockchain: The reason I am explicitly having to define an ECPair fromPrivateKey,  is because bip32 class no longer exposes a keypair attribute.

This portion of the code is broken (I had not tested this with second password).

It should look like the if statement after it.

```
.map(xp => {
  const node = derivePrivateKey(network, xp, chain, index)
  return Bitcoin.ECPair.fromPrivateKey(node.privateKey).toWIF()
})
```

packages/blockchain-wallet-v4/src/types/KVStoreEntry.js
mpfluger: I'm sure there was some reason, but I can't remember. Maybe @xenog does?

packages/blockchain-wallet-v4/src/utils/btc.js
mpfluger: OK cool, if that's the case, def the right thing to do here

packages/blockchain-wallet-v4/src/utils/btc.js
seanblockchain: https://github.com/bitcoinjs/bitcoinjs-lib/blob/master/src/ecpair.js

Screen Shot 2019-11-04 at 4 04 26 PM

This breaking makes sense since the the new ecpair export is really just a set of functions and not the ECPair class itself. 

https://github.com/bitcoinjs/bitcoinjs-lib/blob/master/src/ecpair.js

Will look for a better way to check for ECPair class.

seanblockchain:
```
export const isKey = btcKey => {
  const mockKey = ECPair.fromPrivateKey(crypto.sha256('mock privatekey'))
  return btcKey.constructor === mockKey.constructor
}
```

packages/blockchain-wallet-v4/package.json
```diff
@@ -16,7 +16,7 @@
     "coverage": "cross-env ./../../node_modules/.bin/jest --coverage",
     "link:resolved:paths": "ln -sf $(pwd)/src/** ./node_modules && ln -sf $(pwd)/../../packages/blockchain-wallet-v4-frontend ./node_modules",
     "start": "babel-node src/index.js",
-    "test": "cross-env ./../../node_modules/.bin/jest --silent",
+    "test": "cross-env ./../../node_modules/.bin/jest",
```
plondon: @schnogz is --silent for a reason?

PR #1943, opened 2019-11-07
fix(captcha): refresh captcha if entered wrong

jjBlockchain:
WEB4-1288


schnogz:
waiting for backend to be released to prod

PR #1994, opened 2019-11-21
fix(translations): reset 2fa

schnogz:


plondon:
I'm really not sure what is going on here, span should work

PR #1998, opened 2019-11-22
Feat(bch address utils)

seanblockchain:
## Imported BCH Address  Utils:
1. show private key
2. sign message
3. bip38 imports
4. bip38 sends
5. verify message
6. transfer all
7. send from watch-only
7. accepting both normal and cash addr formatted inputs
8. update wallet context to reference bch  kv-store
9. modify address attributes from kv-store

https://docs.google.com/document/d/1cPfMcWcEsomSoMBOw2c27JyK2uv0kfkEg4RpfzWJ_eY/edit?usp=sharing


packages/blockchain-wallet-v4-frontend/src/modals/Bch/VerifyMessageBch/services.js
```diff
@@ -0,0 +1,17 @@
...
+
+export const verifySignature = ({ address, message, signature }) => {
+  // Exceptions are thrown if the signature is malformed or doesn't match the
+  // address.
+  try {
+    // first need to convert back from cash addr
+    const addr = fromCashAddr(address)
+    return BitcoinMessage.verify(message, addr, signature)
+  } catch (exception) {
+    return false
```
plondon: Maybe return the actual error and display to user?

plondon:
Screen Shot 2019-11-25 at 5 10 19 PM

Verify Message from BCH says Bitcoin, could just use `Address` for the form field label.

plondon:
```Uncaught TypeError: Cannot read property 'final_balance' of undefined```

When clicking More Options -> Private Key

plondon:
Editting a label edits the wallet payload instead of kvStore, is that intended?

seanblockchain:
@plondon thanks for looking this over.

> Editing a label edits the wallet payload instead of kvStore, is that intended?

For now yes. After cleaning up this code, going to move to updating to kv store.

> Uncaught TypeError: Cannot read property 'final_balance' of undefined

Show priv is working on my end. What type of imported address was this? (watch-only, bip38, ...)

seanblockchain:
> Maybe return the actual error and display to user?

Bitcoin cash address field is validated separately in the form.

https://github.com/bitcoinjs/bitcoinjs-message/blob/master/index.js#L148

The bitocinjs verify message util simply returns true or false. It only throws if the signature length is not correct.

seanblockchain:
Everything now runs off of the kv-store.

PR #2055, opened 2019-12-31
Feat/typescript

plondon:
Initial setup of typescript support for the wallet-frontend.
So far this is just allowing js and ts to work together, standards and guidelines to follow.


packages/blockchain-wallet-v4-frontend/src/scenes/Airdrops/StxAirdrop/model.tsx
```diff
@@ -27,10 +28,10 @@ const BlueCartridge = styled(CustomCartridge)`
 // userCampaignState: "NONE", "REGISTERED", "TASK_FINISHED", "REWARD_SEND", "REWARD_RECEIVED", "FAILED"
 // userCampaignTransactionResponseList: []
 export const StxStatus = ({
-  userCampaignsInfoResponseList = [],
+  userCampaignsInfoResponseList,
```
schnogz: do we not need to default to an array anymore?

plondon: Nope, because of line 61

PR #2101, opened 2020-01-22
Feat(typescript): Added typing to the http service

richardwardza:
## Description (optional)
Added typescript support to the http library.

I left the types local to the file as I do not see them being used elsewhere - they are very specific to `axios` and the request methods.

## Testing Steps (optional)
Ensure the existing test suite passes.


packages/blockchain-wallet-v4/src/network/api/http.ts
```diff
@@ -0,0 +1,100 @@
...
+  ignoreQueryParams: boolean
+}
+
+type Header = {
+  Authorization?: string
+  ['Content-Type']: string
+}
+
+export default ({ apiKey }: { apiKey: string }) => {
+  const encodeData = (data: any, contentType: string) => {
```
richardwardza: I've set `data` to `any` as that is how it's defined in `AxiosRequestConfig`.
I'm unsure what would be sent through but it could possibly be updated to something like `[k:string]: string | number | object`

schnogz: I think `any` is fine for now. I cant recall off the top of my head if we even use this field beyond the defaults. @plondon do you?

schnogz:
🎉

plondon:
Hey Richard, thanks for the PR!

I think overall it's great, the main problem that I see is that when `get` or `post` or any of these methods is used the types are lost. So we need to do something about that, otherwise 👍

PR #2102, opened 2020-01-23
Feat/typescript exchange

TheLeoB:
## Description (optional)
Tested locally, exchange flow works. 

## Testing Steps (optional)


schnogz:
@plondon how's this look?

packages/blockchain-wallet-v4-frontend/src/data/components/exchange/reducers.ts
```diff
@@ -0,0 +1,79 @@
+import * as AT from './actionTypes'
+import { ExchangeActionTypes, ExchangeState } from './types'
+import Remote from 'blockchain-wallet-v4/src/remote/remote'
+
+const INITIAL_STATE: ExchangeState = {
+  limits: Remote.NotAsked,
+  min: null,
+  max: null,
+  // @PHIL how to solve for default state for all these, 15-28
```
plondon: @TheLeoB I don't think the INITIAL_STATE should change from https://github.com/blockchain/blockchain-wallet-v4-frontend/pull/2102/files#diff-80ce3fa62bee1d23850e68184db9e267L5-L13.

If the value can be null we should reflect that in `ExchangeState ` here https://github.com/blockchain/blockchain-wallet-v4-frontend/pull/2102/files#diff-3c02af347613331721c9487161a17413R37-R49

packages/blockchain-wallet-v4-frontend/src/data/components/exchange/types.ts
```diff
@@ -34,20 +34,25 @@ export type LimitDurationType = {
...
+          min: number
+        }
+        priority: number
+        regular: number
+      }
+      source: number | string
+      sourceFiat: string
+      target: number | string
+    }
+  | {
```
TheLeoB: I added this to source fee type. Everything worked, was able to make an exchange.

packages/blockchain-wallet-v4-frontend/src/data/components/exchange/reducers.ts
TheLeoB: Made the change to types file. targetFee was removed from initial state since it wasn't being used.

PR #2125, opened 2020-02-21
Feat/borrow changes

plondon:
## Description (optional)

## Testing Steps (optional)

### DIFF
--- a/packages/blockchain-wallet-v4-frontend/src/scenes/Borrow/index.tsx
+++ b/packages/blockchain-wallet-v4-frontend/src/scenes/Borrow/index.tsx
@@ -45,11 +45,23 @@ class Borrow extends PureComponent<Props, State> {
     }
 
     this.props.borrowActions.fetchBorrowOffers()
+    this.checkUserData()
+  }
 
+  componentDidUpdate (prevProps: Props) {
+    if (
+      this.props.userDataR.getOrElse(null) !==
+      prevProps.userDataR.getOrElse(null)
+    ) {
+      this.checkUserData()
+    }
+  }
+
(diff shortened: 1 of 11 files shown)


packages/blockchain-wallet-v4-frontend/src/scenes/Borrow/index.tsx
```diff
@@ -32,14 +37,24 @@ export const MainTitle = styled(Text)`
...
 
     this.props.borrowActions.fetchBorrowOffers()
+
+    const userData = this.props.userDataR.getOrElse({
+      tiers: { current: 0 }
+    })
+    const tier = userData.tiers.current
+    const isDisabled = tier < 2
+    /* eslint-disable */
+    this.setState({ isDisabled })
```
schnogz: is there a way around this? this is going to always cause at least a double render

plondon: > You may call setState() immediately in componentDidMount(). It will trigger an extra rendering, but it will happen before the browser updates the screen. This guarantees that even though the render() will be called twice in this case, the user won’t see the intermediate state. Use this pattern with caution because it often causes performance issues. In most cases, you should be able to assign the initial state in the constructor() instead. It can, however, be necessary for cases like modals and tooltips when you need to measure a DOM node before rendering something that depends on its size or position.

plondon: We can start state with isDisabled: true instead of false, that way rerender wont happen unless its supposed to.

PR #2126, opened 2020-02-21
Feat(input redesign)

jjBlockchain:


packages/blockchain-wallet-v4-frontend/src/scenes/Register/template.js
```diff
@@ -155,6 +156,7 @@ const Register = ({
                 data-e2e='signupPassword'
                 name='password'
                 validate={[required, validStrongPassword]}
+                noLastPass
```
plondon: What if the user uses lastpass to generate a password?

packages/blockchain-info-components/src/Form/TextInput.js
```diff
@@ -98,15 +130,36 @@ class TextInput extends React.Component {
   }
 
   render () {
-    const { disabled, errorState, icon, iconSize, ...rest } = this.props
+    const {
+      controlledBgColor,
+      controlledBorderColor,
+      controlledFocusBorderColor,
```
plondon: I don't see this being used anywhere

packages/blockchain-info-components/src/Form/PasswordInput.js
```diff
@@ -61,16 +95,32 @@ class PasswordInput extends React.Component {
   }
 
   render () {
-    const { errorState, ...rest } = this.props
-    const borderColor = selectBorderColor(
-      this.props.controlledBorderColor || errorState
+    const {
+      active,
+      controlledBgColor,
+      controlledBorderColor,
```
plondon: It would be good to clean this up, if we can be consistent with all inputs and don't need a `controlledBorderColor` override lets completely remove this.

plondon:
Screen Shot 2020-02-23 at 12 01 11 PM

Password bar border radius issue

plondon:
Screen Shot 2020-02-23 at 12 02 26 PM

Only NumberInput has the background and border treatments. We need to update SelectBoxes and Textarea as well.

plondon:
Screen Shot 2020-02-23 at 12 06 29 PM

Password stretching form needs treatment

plondon:
Screen Shot 2020-02-23 at 12 07 06 PM

Autologout needs treatment

packages/blockchain-info-components/src/Colors/Default.ts
```diff
@@ -50,7 +50,7 @@ const Default = {
   orange700: '#D96716',
   orange800: '#B24400',
   orange900: '#8C2F00',
-  red000: '#FAECEB',
+  red000: '#FFECEB',
```
plondon: https://github.com/blockchain/components/blob/master/src/Tools/palette.js#L41

Did you mean to change this?

jjBlockchain: https://www.figma.com/file/Nks0uPvdzq4hOTPkvuebhK/Inputs?node-id=0%3A1 <--- looks like design changed red000?

packages/blockchain-wallet-v4-frontend/src/scenes/Register/template.js
jjBlockchain: I was trying to find a way to detect whether or not users has lp as an extension, but it looks like I wasn't able to find a way so i will remove noLastPass for now

PR #2144, opened 2020-03-06
Fix(Loans): Dont close loan if collateral payment fails

schnogz:
@plondon feel free to modify as you see fit


plondon:
We still should call notify deposit and tell the user that their transaction failed and the loan is not yet closed.

plondon:
I'm going to handle this in the transactions PR

PR #2145, opened 2020-03-07
Fix/loan txs

plondon:
## Description (optional)
Display actionable info when a loan deposit or repayment fails. Need to wait for timestamp from backend for proper sorting of txs so we know the most recent

## Testing Steps (optional)


packages/blockchain-wallet-v4-frontend/src/data/components/borrow/model.ts
```diff
@@ -53,3 +54,25 @@ export const fiatDisplayName = (coin: CoinType) => {
       return 'USD'
   }
 }
+
+export const lastTxFailed = (
+  loan: LoanType,
+  loanTransactions: Array<LoanTransactionsType>
+): boolean => {
+  const lastTx = head(loanTransactions)
```
plondon: This may need to change

PR #2260, opened 2020-04-13
Feat/recovery design

TheLeoB:
## Testing Steps (optional)


packages/blockchain-wallet-v4-frontend/src/data/components/recoveryPhrase/reducers.ts
```diff
@@ -0,0 +1,33 @@
...
+          ...state,
+          step: action.payload.step
+        }
+      default: {
+        return {
+          ...state,
+          step: action.payload.step
+        }
+      }
+    }
```
plondon, line 29: For borrow and sb there were other state props we were changing depending on the step. I don't think you need lines 14-23, you can just change the step.

TheLeoB:
Ready for final review/merge

PR #2267, opened 2020-04-16
Feat/savings flyout

jjBlockchain:
* Changed checkbox to flex-start instead of vertically center
* Added flyout component for savings deposit
* Relocated BorrowDropdown to component level and renamed to CoinBalanceDropdown


packages/blockchain-wallet-v4-frontend/src/data/components/interest/sagas.ts
```diff
@@ -0,0 +1,35 @@
...
+      case 'BTC':
+        const accountsR = yield select(
+          selectors.core.common.btc.getAccountsBalances
+        )
+        const defaultIndex = yield select(
+          selectors.core.wallet.getDefaultAccountIndex
+        )
+        defaultAccountR = accountsR.map(nth(defaultIndex))
+        break
+    }
```
plondon, line 24: Might want to throw an error for other coins if their not set up yet.

PR #2290, opened 2020-04-23
Fix(cleanup console)

schnogz:
- removes logging artifacts from developing
- ensure matomo iframe is loaded before posting message to avoid console errors

Screen Shot 2020-04-23 at 2 14 15 PM


plondon:
Nice, thank you!

PR #2314, opened 2020-05-01
fix(recovery-phrase-data): add data-e2e tag

jjBlockchain:


packages/blockchain-wallet-v4-frontend/src/modals/RecoveryPhrase/ShowRecoveryWords/template.tsx
```diff
@@ -80,7 +80,7 @@ const WordsList = ({ handleBackArrow, handleNextButton, step, words }) => {
           if (index >= 6) {
             return (
               <WordBox>
-                <WordText>
+                <WordText data-e2e='backupWords'>
```
schnogz, line 83: these need to be unique otherwise how are they doing to target them... 
`data-e2e={``backupWord${index}``}`

schnogz: nevermind i see whats going on here. this is okay 👍

PR #2389, opened 2020-05-12
fix(rates): revert back to origin rate type

TheLeoB:
## Description (optional)
response from savings/rates is now formatted as it was originally. 

## Testing Steps (optional)


packages/blockchain-wallet-v4/src/network/api/interest/types.ts
```diff
@@ -31,10 +31,7 @@ export type InterestAccountType = {
 }
 
 export type InterestRateType = {
-  rates: {
-    BTC: number
-    ETH: number
-  }
+  BTC: number
```
plondon, line 34: Do you always only have BTC?

PR #2499, opened 2020-06-29
Fix/balance borrow

TheLeoB:
## Description (optional)
Added a selector file that just gets nonCustodial (HD wallets only for BTC/BCH, no imported addresses). Only being used by borrow so far, to display initial estimates of how much user can borrow. 

## Testing Steps (optional)


packages/blockchain-wallet-v4-frontend/src/components/Balances/nonCustodial/selectors.ts
```diff
@@ -0,0 +1,104 @@
...
+    case 'PAX':
+      return getPaxBalance
+    case 'XLM':
+      return getXlmBalance
+    case 'USDT':
+      return getUsdtBalance
+    default:
+      return Remote.Failure(INVALID_COIN_TYPE)
+  }
+}
```
milan-bc, line 104: Suggestion: we might make combine selector function for PAX/USDT like

```
export const getMiscBalance = (coin) => createDeepEqualSelector(
  [state => selectors.core.data.eth.getErc20Balance(state, coin)],
  balanceR => {
    return Remote.of(new BigNumber(balanceR.getOrElse(0)))
  }
)
```

and then in switch 

```

export const getBalanceSelector = (coin: CoinType) => {
  switch (coin) {
    case 'BTC':
      return getBtcBalance
    case 'BCH':
      return getBchBalance
    case 'ETH':
      return getEthBalance
    case 'XLM':
      return getXlmBalance
    case 'PAX':
    case 'USDT':
      return getMiscBalance(toLower(coin))
    default:
      return Remote.Failure(INVALID_COIN_TYPE)
  }
}
```

TheLeoB: Thanks for this, I made this change. We could also do this for all balance selectors in the wallet, but will tackle separately, since it will have a larger impact on the wallet.

milan-bc: yeah that's a plan so far :)

plondon: > We could also do this for all balance selectors in the wallet, but will tackle separately, since it will have a larger impact on the wallet.

I think this should be done from the algo/usdt branches?

PR #2503, opened 2020-06-30
feat: fixed propagation of proper coin name

milan-bc:
## Description (optional)
This PR solve issue for propagating proper coin to selector and then to run transaction call 

## Testing Steps (optional)
login to wallet 
try to send some usdt
make sure that there are no errors in network tab


plondon:
Please add typescript to these files

PR #2509, opened 2020-07-03
feat: implemented select crypto step

milan-bc:
## Description (optional)
Moved stuff to be in new step for choosing a crypto value

## Testing Steps (optional)
Go to Buy Crypto, make sure that new step is visible


packages/blockchain-wallet-v4-frontend/src/data/components/simpleBuy/actions.ts
```diff
@@ -418,14 +418,29 @@ export const setStep = (
         cryptoCurrency?: CoinType
         defaultMethod?: SBFormPaymentMethod
         fiatCurrency: FiatType
+        pair: SBPairType
         step: 'ENTER_AMOUNT'
       }
+    | {
+        cryptoCurrency?: CoinType
+        defaultMethod?: SBFormPaymentMethod
```
plondon, line 426: I don't think you would have a defaultMethod yet?

PR #2515, opened 2020-07-11
RFC: Fix/lift typings

plondon:
## Description (optional)
As described on slack, there are some limitations to typing ramda's `lift` function. This is what I have so far, it still requires you to give the "`transform`ation" function some information about your Remote types, but some of the heavy lifting is now done by using Typescript's `extend`  and `infer` to get the `SuccessStateType`.

**Note:** This PR breaks getOrElse, getOrFail, and mapped, which was previously incorrectly typed.

### What improved?
I was able to get rid of `SuccessStateType` and `LinkStatePropsType`:

Screen Shot 2020-07-11 at 12 56 51 PM

These are now moved to the selectors.ts file, but we still need to use the new `ExtractSuccess` when defining the types of the params for the "`transform`ation" function.

Screen Shot 2020-07-11 at 1 12 03 PM

Now getData knows the return type of our `lift` w/o needing to explicitly set the `SuccessStateType`'s

Screen Shot 2020-07-11 at 1 13 55 PM

### What can be improved more?
My biggest frustration is that I can not figure out how to get the Remote's `Success` type inside the "`transform`ation" function without defining it explicitly. As you can see [here](https://github.com/blockchain/blockchain-wallet-v4-frontend/compare/fix/lift-typings?expand=1#diff-dfdd32de94cc282e5b0a19d9627f83bbR7) I had to use `any[]` for `args`. In an ideal world `any[]` would be `Array<ExtractSuccess[A]>`, but A is not defined until the next function below...

## Testing Steps (optional)

### DIFF
--- a/packages/blockchain-wallet-v4-frontend/src/modals/SimpleBuy/EnterAmount/index.tsx
+++ b/packages/blockchain-wallet-v4-frontend/src/modals/SimpleBuy/EnterAmount/index.tsx
@@ -1,14 +1,6 @@
 import { actions, selectors } from 'data'
 import { bindActionCreators, Dispatch } from 'redux'
 import { connect, ConnectedProps } from 'react-redux'
-import {
-  FiatEligibleType,
-  FiatType,
-  RemoteDataType,
-  SBCardType,
-  SBPairType,
-  SBPaymentMethodsType
-} from 'core/types'
 import { getData } from './selectors'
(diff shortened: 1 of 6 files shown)


plondon:
Closing this, we can reuse some things from it but for the most part this will be more work in the long run since we **need** SuccessStateType anyway...

PR #2527, opened 2020-07-22
feat: deposit screen implemented wall-7

milan-bc:
## Description (optional)
this PR implements deposit screen

## Testing Steps (optional)
Go over buy/sell flow, click on deposit and make sure that you see new deposit step


packages/blockchain-wallet-v4-frontend/src/components/SimpleBuy/MultiRowContainer.tsx
```diff
@@ -0,0 +1,12 @@
+import styled from 'styled-components'
+
+const MultiRowContainer = styled.div`
+  position: relative;
+  display: flex;
+  flex-direction: column;
+  width: 230px;
```
plondon, line 7: Why hardcode?

packages/blockchain-wallet-v4-frontend/src/modals/SimpleBuy/OrderSummary/template.success.tsx
```diff
@@ -93,7 +93,7 @@ const Success: React.FC<Props> = props => {
           <div style={{ margin: '16px 0' }}>
             <Status {...props} />
           </div>
-          {props.order.state === 'PENDING_DEPOSIT' &&
+          {/* {props.order.state === 'PENDING_DEPOSIT' &&
```
plondon, line 96: No more orders so we can delete this I think

packages/blockchain-wallet-v4-frontend/src/modals/SimpleBuy/PaymentMethods/Payments/index.tsx
```diff
@@ -97,7 +97,13 @@ class Payments extends PureComponent<InjectedFormProps<{}, Props> & Props> {
           />
         )
       case 'FUNDS':
-        return <></>
+        return (
+          <Icon
+            size='32px'
+            color='fiat'
+            name={value.currency === 'EUR' ? 'eur' : 'gbp'}
```
plondon, line 104: value.currency.toLowerCase()

milan-bc: nope that doesn't work due to typing...

packages/blockchain-wallet-v4-frontend/src/components/SimpleBuy/MultiRowContainer.tsx
milan-bc: for better alignment, will re-check about 100%

packages/blockchain-wallet-v4-frontend/src/modals/SimpleBuy/PaymentMethods/Payments/index.tsx
plondon: value.currency.toLowerCase() as key of DefaultTheme

plondon:
Some style/data tweaks:

This amount should be the balance not the limit:

Screen Shot 2020-07-22 at 4 48 24 PM

Oval: 

Screen Shot 2020-07-22 at 4 44 35 PM

Missing: 

Screen Shot 2020-07-22 at 4 43 33 PM

plondon:
Can we align the x and the copy?

Screen Shot 2020-07-22 at 4 43 17 PM

packages/blockchain-wallet-v4-frontend/src/modals/SimpleBuy/PaymentMethods/Payments/index.tsx
milan-bc: we had to make it something like `name={value.currency.toLowerCase() as 'eur' | 'gbp'}`

milan-bc:
@plondon nope we can't align them because per design they are not aligned X and copy buttons...

plondon:
They're aligned in all other steps but not in designs...I'll check w/ Greg

packages/blockchain-wallet-v4-frontend/src/components/CopyClipboardButton/index.tsx
```diff
@@ -0,0 +1,58 @@
...
+        handleClick={this.handleClick}
+        data-e2e={this.props['data-e2e']}
+      />
+    )
+  }
+}
+
+const mapDispatchToProps = dispatch => ({
+  alertActions: bindActionCreators(actions.alerts, dispatch)
+})
```
plondon, line 56: const connector = connect(null, mapDispatchToProps)

type Props = ConnectedProps<typeof connector>

packages/blockchain-wallet-v4-frontend/src/components/CopyClipboardButton/index.tsx
```diff
@@ -0,0 +1,59 @@
+import * as C from 'services/AlertService'
+import { actions } from 'data'
+import { bindActionCreators } from 'redux'
+import { connect, ConnectedProps } from 'react-redux'
+import CopyClipboard from './template'
+import React from 'react'
+
+export interface OwnProps {
+  address: string
+  alertActions: any
```
plondon, line 10: Can you change `address` => `text`?

If you're using ConnectedProps<typeof connector> you don't need to set alertActions here. I would consider `OwnProps` as the props that are passed in. `alertActions` is connected via mapDispatchToProps, so it's not passed in from a parent.

packages/blockchain-wallet-v4-frontend/src/components/CopyClipboardButton/index.tsx
```diff
@@ -0,0 +1,59 @@
...
+}
+
+export interface State {
+  active: boolean
+}
```
plondon, line 15: I don't think it really matters if we use `interface` or `type` for props, but since we use `type` everywhere else I would urge you to use that for the sake of consistency. We also tend place types below the component because they depend on `const connector` and this way they are all together. 👇

packages/blockchain-wallet-v4-frontend/src/components/CopyClipboardButton/template.tsx
```diff
@@ -26,11 +27,10 @@ const CopyButton = styled(IconButton)<CopyButtonProps>`
   }
 `
 
-interface CopyClipboardProps {
-  active: boolean
-  address: string
-  handleClick: () => void
-}
+type CopyClipboardProps = OwnProps['alertActions'] &
```
plondon, line 30: Is alertActions used in this component?

PR #2561, opened 2020-08-02
fix(int.): fix deposit payment handling

plondon:
## Description (optional)

## Testing Steps (optional)

### DIFF
--- a/packages/blockchain-wallet-v4-frontend/src/data/components/exchange/sagas.utils.ts
+++ b/packages/blockchain-wallet-v4-frontend/src/data/components/exchange/sagas.utils.ts
@@ -88,6 +88,8 @@ export default ({ coreSagas, networks }) => {
         .build()
         .done()).value()
     } catch (e) {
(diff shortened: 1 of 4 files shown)


packages/blockchain-wallet-v4-frontend/src/data/components/exchange/sagas.utils.ts
```diff
@@ -88,6 +88,8 @@ export default ({ coreSagas, networks }) => {
         .build()
         .done()).value()
     } catch (e) {
+      // eslint-disable-next-line
+      console.log(e)
```
TheLeoB, line 92: Why'd you want to log the error here?

plondon: This usually ends up being where an error happens and it goes unnoticed because we `return {}`. It would probably be better to actually throw the error but I dont want to break existing code somewhere else :/

PR #2637, opened 2020-09-08
Fix(language locale): fixed problem with empty language

milan-bc:
## Description (optional)
This PR solve problem with react-intl massive errors in console

## Testing Steps (optional)

### DIFF
--- a/packages/blockchain-wallet-v4-frontend/src/providers/TranslationsProvider/index.js
+++ b/packages/blockchain-wallet-v4-frontend/src/providers/TranslationsProvider/index.js
@@ -8,7 +8,7 @@ import { selectors } from 'data'
 
 class TranslationsProvider extends React.Component {
   state = {
-    locale: '',
+    locale: 'en',
     messages: {}
   }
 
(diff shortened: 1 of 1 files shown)


packages/blockchain-wallet-v4-frontend/src/providers/TranslationsProvider/index.js
```diff
@@ -8,7 +8,7 @@ import { selectors } from 'data'
 
 class TranslationsProvider extends React.Component {
   state = {
-    locale: '',
+    locale: 'en',
```
milan-bc, line 11: setting this to be default language, we have init function which is not called at all.
We do call this [saga](https://github.com/blockchain/blockchain-wallet-v4-frontend/blob/development/packages/blockchain-wallet-v4-frontend/src/data/rootSaga.js#L41-L54) after we tried to get react-intl stuff. 

However react-intl have some upgrades to version 5, we are still using v2, it might be worth to migrate to latest, I opened a tech debt ticket for it

schnogz (approved):

PR #2679, opened 2020-09-23
feat(qr read): we show spinner once when user read QR code

milan-bc:
## Description (optional)
This PR adds loader once when user scan QR code on mobile logic

## Testing Steps (optional)


schnogz (approved):
Awesome work! This bug has been around for far too long.  Going to test it out now before merging

PR #2701, opened 2020-10-02
Feat(Interest): add BCH and XLM support

schnogz:


config/mocks/wallet-options-v4.json
```diff
@@ -195,7 +195,7 @@
           "colorCode": "xlm",
           "config": {
             "network": "public",
-            "sendTimeOutSeconds": 10
+            "sendTimeOutSeconds": 600
```
schnogz, line 198: now matches prod wallet options

PR #2729, opened 2020-10-16
feat(ssd flow): implemented use of currency based on PI address and country code

milan-bc:
## Description (optional)

## Testing Steps (optional)


packages/blockchain-wallet-v4-frontend/src/data/auth/sagas.js
```diff
@@ -127,6 +128,13 @@ export default ({ api, coreSagas }) => {
       yield call(coreSagas.data.xlm.fetchLedgerDetails)
       yield call(coreSagas.data.xlm.fetchData)
 
+      const countryCodeR = yield select(selectors.core.settings.getCountryCode)
+      const countryCode = countryCodeR.getOrElse('US')
+      const currency = getCurrencyBasedOnCountry(countryCode)
+      yield put(actions.preferences.setSBFiatCurrency(currency))
+      // TODO - update wallet with currency should be done on BE side
+      yield put(actions.core.settings.setCurrency(currency))
```
milan-bc, line 136: I see this as techDebt reason is there we have to set it back to wallet because otherwise it will be overwritten with https://github.com/blockchain/blockchain-wallet-v4-frontend/blob/development/packages/blockchain-wallet-v4-frontend/src/data/preferences/sagas.ts#L24-L38

plondon: @milan-bc I think we should only run this if `firstLogin` is true

plondon: Can we just remove sbFiatCurrency and assume from now on that your sbFiatCurrency === walletCurrency?

packages/blockchain-wallet-v4-frontend/src/data/components/simpleBuy/sagas.ts
```diff
@@ -853,38 +853,38 @@ export default ({
...
-          step: 'ENTER_AMOUNT',
-          cryptoCurrency,
-          fiatCurrency
-        })
-      )
-    } else {
-      yield put(
-        A.setStep({ step: 'CRYPTO_SELECTION', cryptoCurrency, fiatCurrency })
-      )
+    if (fiatCurrency) {
```
plondon, line 856: Do we need fiatCurrency?

PR #2730, opened 2020-10-21
fix(terms): wrap accept terms set in useeffect

plondon:
## Description (optional)

## Testing Steps (optional)

### DIFF
--- a/packages/blockchain-wallet-v4-frontend/src/modals/SimpleBuy/CheckoutConfirm/template.success.tsx
+++ b/packages/blockchain-wallet-v4-frontend/src/modals/SimpleBuy/CheckoutConfirm/template.success.tsx
@@ -22,7 +22,7 @@ import { Props as OwnProps, SuccessStateType } from '.'
 import { SupportedWalletCurrenciesType } from 'core/types'
 
 import { displayFiat, getPaymentMethod } from '../model'
-import React, { useState } from 'react'
+import React, { useEffect, useState } from 'react'
 import styled from 'styled-components'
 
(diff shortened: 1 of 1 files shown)


milan-bc (approved):

packages/blockchain-wallet-v4-frontend/src/modals/SimpleBuy/CheckoutConfirm/template.success.tsx
```diff
@@ -70,15 +70,15 @@ const Success: React.FC<InjectedFormProps<{}, Props> & Props> = props => {
...
-      props.order.paymentType === 'USER_CARD'
-    )
-  ) {
-    setAcceptTerms(true)
-  }
+  useEffect(() => {
+    if (!requiresTerms) {
+      return setAcceptTerms(true)
+    }
+  }, [])
```
milan-bc, line 81: here we can pass `acceptTerms` and then it will run useEffect only in case it has been changed

plondon: Why would we need to do that?

milan-bc: that way we said to react skip applying it unless it has been changed

milan-bc: but not mandatory at all, just a note

PR #2748, opened 2020-11-02
feat(banners): load user id in case that user is not in kvstore

milan-bc:
## Description (optional)

## Testing Steps (optional)

### DIFF
--- a/packages/blockchain-wallet-v4-frontend/src/data/analytics/sagas.ts
+++ b/packages/blockchain-wallet-v4-frontend/src/data/analytics/sagas.ts
@@ -226,6 +226,7 @@ export default () => {
     initUserSession,
     postMessage,
     startSession,
-    stopSession
+    stopSession,
+    waitForUserId
   }
 }
--- a/packages/blockchain-wallet-v4-frontend/src/data/analytics/sagas.ts
+++ b/packages/blockchain-wallet-v4-frontend/src/data/analytics/sagas.ts
@@ -20,7 +20,7 @@ export default ({
   coreSagas: any
   networks: any
 }) => {
-  const { waitForUserId } = profileSagas({
+  const { waitForUserData } = profileSagas({
     api,
     coreSagas,
     networks
@@ -50,7 +50,7 @@ export default ({
     const defaultHDWallet = yield select(
       selectors.core.wallet.getDefaultHDWallet
     )
-    const userId = yield call(waitForUserId)
+    const userId = yield call(waitForUserData)
     if (userId) return userId
     const { seedHex } = defaultHDWallet
     const mnemonic = BIP39.entropyToMnemonic(seedHex)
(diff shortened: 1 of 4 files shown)


packages/blockchain-wallet-v4-frontend/src/data/analytics/sagas.ts
```diff
@@ -226,6 +228,7 @@ export default () => {
     initUserSession,
     postMessage,
     startSession,
-    stopSession
+    stopSession,
+    waitForUserId
```
plondon, line 232: @milan-bc probably don't need to export an imported function

packages/blockchain-wallet-v4-frontend/src/data/modules/profile/sagas.ts
```diff
@@ -554,6 +565,7 @@ export default ({ api, coreSagas, networks }) => {
     syncUserWithWallet,
     updateUser,
     updateUserAddress,
-    waitForUserData
+    waitForUserData,
+    waitForUserId
```
plondon, line 569: Let's not export waitForUserId, let's only expose waitForUserData. In analytics saga lets also just waitForUserData

packages/blockchain-wallet-v4-frontend/src/data/modules/profile/sagas.ts
```diff
@@ -29,10 +29,22 @@ export const renewUserDelay = 30000
...
-  const waitForUserData = function * () {
-    const userId = (yield select(
+  const waitForUserId = function * () {
+    const userId = yield select(
+      selectors.core.kvStore.userCredentials.getUserId
+    )
+    if (Remote.Success.is(userId)) return userId.getOrElse(null)
+    yield take(
+      actionTypes.core.kvStore.userCredentials
+        .FETCH_METADATA_USER_CREDENTIALS_SUCCESS
```
plondon, line 39: If user credentials metadata fails to load (`FETCH_METADATA_USER_CREDENTIALS_FAILURE `) then this take will never resolve.

packages/blockchain-wallet-v4-frontend/src/data/analytics/sagas.ts
milan-bc: yes, we do not need it

packages/blockchain-wallet-v4-frontend/src/data/modules/profile/sagas.ts
milan-bc: @plondon should we trow an error in case of failure? I think it make sense

milan-bc: actually never mind will apply same approach we do in waitForUserData

plondon (approved):

PR #2755, opened 2020-11-06
feat(swap-2.x): min clarification and notes

plondon:
## Description (optional)
Min and Max swap (just min right now)

## Testing Steps (optional)

### DIFF
--- a/packages/blockchain-wallet-v4-frontend/src/modals/Swap/EnterAmount/Checkout/validation.tsx
+++ b/packages/blockchain-wallet-v4-frontend/src/modals/Swap/EnterAmount/Checkout/validation.tsx
@@ -69,14 +69,8 @@ export const getMaxMin = (
       // 4517 satoshi is 0.017672510 ETH is 7 USD (AOTW)
       const counterFeeInBase = exRate.times(standardCounterFee).toNumber()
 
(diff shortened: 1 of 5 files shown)


packages/blockchain-wallet-v4-frontend/src/modals/Swap/EnterAmount/Checkout/validation.tsx
```diff
@@ -30,34 +30,53 @@ export const getMaxMin = (
...
-      const counterMin = counterRate
-        .times(standardCounterFee)
-        .plus(staticFee)
-        .toNumber()
+      // 4517 satoshi is 0.017672510 ETH is 7 USD (AOTW)
+      const counterFeeInBase = exRate.times(standardCounterFee).toNumber()
 
-      return Math.max(counterMin, baseMin).toPrecision(CRYPTO_DECIMALS)
+      // .017672510 ETH (7 USD) is greater than .01123091 ETH (3.5 USD)
+      if (counterFeeInBase > baseMin) {
```
niall-bc, line 73: I don't think we need this condition, a better UX would be to always return counterFeeInBase + baseMin, as this wouldn't fix the case where baseMin = 1.0 and counterFeein base = 0.99 for example

PR #2775, opened 2020-11-17
Fix/eligibility check

TheLeoB:
## Description (optional)
https://blockchain.atlassian.net/browse/FWLT-847

Add eligibility check to `/eligible/product/swap` to see if user is able to swap to/from custodial wallets. If eligible: false, user only sees non-custodial wallets as an option for swapping.


packages/blockchain-wallet-v4-frontend/src/modals/Swap/CoinSelection/index.tsx
```diff
@@ -85,8 +87,19 @@ class CoinSelection extends PureComponent<Props> {
...
 
+  checkCustodialEligibility = (
+    custodialEligibility: boolean,
+    account: SwapAccountType
+  ) => {
+    if (account.type === 'CUSTODIAL' && !custodialEligibility) {
+      return false
+    } else {
+      return true
+    }
```
schnogz, line 98: this could be simplified to just 
`return !(account.type === 'CUSTODIAL' && !custodialEligibility)`

schnogz (approved):
Really nice work. Very clean 👏

milan-bc (approved):

PR #2778, opened 2020-11-18
fix(swap): btc fee amount displayed in preview swap screen

TheLeoB:
For btc/bch outgoing swaps, we display the network fee as the sat/b value rather than what the network fee actually is. BTC/BCH payment has a network fee under payment.selection.fee, while other coins have a static fee in just payment.fee. 

500A0774-39F3-4D84-843D-1D7CA7DE8403

This should be released to production ASAP but https://github.com/blockchain/blockchain-wallet-v4-frontend/commit/f470e9538dbba591a03769716da99f89327b483d should be cherry picked out of development before release since it breaks swap.

### DIFF
--- a/packages/blockchain-wallet-v4-frontend/src/modals/Swap/PreviewSwap/index.tsx
+++ b/packages/blockchain-wallet-v4-frontend/src/modals/Swap/PreviewSwap/index.tsx
@@ -30,9 +30,11 @@ class PreviewSwap extends PureComponent<InjectedFormProps<{}, Props> & Props> {
   }
 
   networkFee = value => {
-    return value.coin === 'BTC' || value.coin === 'BCH'
-      ? value.selection.fee
-      : value.fee
+    return value
+      ? value.coin === 'BTC' || value.coin === 'BCH'
+        ? value.selection.fee
+        : value.fee
+      : 0
   }
 
   render () {
@@ -174,7 +176,7 @@ class PreviewSwap extends PureComponent<InjectedFormProps<{}, Props> & Props> {
                     {coinToString({
                       value: convertBaseToStandard(
                         BASE.baseCoin,
-                        value ? this.networkFee(value) : 0
+                        this.networkFee(value)
                       ),
                       unit: { symbol: coins[BASE.baseCoin].coinTicker }
                     })}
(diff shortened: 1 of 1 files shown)


schnogz (approved):

packages/blockchain-wallet-v4-frontend/src/modals/Swap/PreviewSwap/index.tsx
```diff
@@ -168,7 +174,7 @@ class PreviewSwap extends PureComponent<InjectedFormProps<{}, Props> & Props> {
                     {coinToString({
                       value: convertBaseToStandard(
                         BASE.baseCoin,
-                        value ? value.fee : 0
+                        value ? this.networkFee(value) : 0
```
blockdylanb, line 177: One small optimization could be to move the `value` check inside of the `networkFee` method so you keep the else logic contained. So it would look like:

```
    value: convertBaseToStandard(
         BASE.baseCoin,
         this.networkFee(value),
    )
```

blockdylanb: not mission critical or anything

TheLeoB: Good callout, moved this out to the method.

packages/blockchain-wallet-v4-frontend/src/modals/Swap/PreviewSwap/index.tsx
```diff
@@ -29,6 +29,14 @@ class PreviewSwap extends PureComponent<InjectedFormProps<{}, Props> & Props> {
     this.props.swapActions.createOrder()
   }
 
+  networkFee = value => {
+    return value
+      ? value.coin === 'BTC' || value.coin === 'BCH'
```
milan-bc, line 34: we should create one global enum for coin names and use it on every place we do have such checks

packages/blockchain-wallet-v4-frontend/src/modals/Swap/PreviewSwap/index.tsx
```diff
@@ -29,6 +29,14 @@ class PreviewSwap extends PureComponent<InjectedFormProps<{}, Props> & Props> {
     this.props.swapActions.createOrder()
   }
 
+  networkFee = value => {
```
milan-bc, line 32: ideally add type for value

fatemehramezanli: تایید

PR #2781, opened 2020-11-20
Feat/sell p3

plondon:
## Description (optional)
Uses swap 2.0 api for sell 3.0. Not an ideal solution for the FE because swap and buy/sell are 2 separate UI flows. A lot of code has been copy pasta'd for the sake of saving time, however, swap quotes are different than simplebuy quotes, and swap orders and different than simplebuy orders. There's definitely some sketchy stuff going on. I tried to leave comments when I made something "sell" specific, but eventually it'll be shared with buy.

```
// used for sell only now, eventually buy as well
// TODO: use swap2 quote for buy AND sell
```

### What still needs to be done?
1. After submitting a sell the account balances do not update

 2. ~The payment fee is the gasPrice or satoshi/byte, not the final fee~
Screen Shot 2020-11-24 at 10 21 55 AM
3. The swap 2.0 API only returns transactions for the swap output currency. So swap's (sells) from BTC-USD are not currently fetched, because fiat transaction lists only fetch custodial (deposits/withdrawal) transactions. The swap 2.0 api fetch transaction should be added to the fiat transaction lists. More reasons why we need 1 tx list...

~4. PreviewSell component does not have an informative Loading state~
Screen Shot 2020-11-24 at 10 35 29 AM
5. `CryptoSelector` component utilizes Swap's `CryptoAccountOption` but they are slightly different (see circular checkmark empty state for example)
Screen Shot 2020-11-24 at 10 38 58 AM
6. Right now we're reusing `SuccessfulSwap` which requires modal `close` of `SB_MODAL` and open of `SWAP_MODAL`. Could easily add a step to SWAP_STEPS and reuse SuccessfulSwap that way.

## Testing Steps (optional)


packages/blockchain-wallet-v4-frontend/src/data/components/simpleBuy/sagas.ts
```diff
@@ -207,20 +235,84 @@ export default ({
...
+      // TODO: use swap2 quote for buy AND sell
+      if (orderType === 'SELL') {
+        const from = S.getSwapAccount(yield select())
+        const quote = S.getSellQuote(yield select()).getOrFail(NO_QUOTE)
+
+        if (!from) throw new Error(NO_ACCOUNT)
+
+        const direction = getDirection(from)
+        const cryptoAmt =
+          fix === 'CRYPTO'
```
plondon, line 263: I hate this

packages/blockchain-wallet-v4-frontend/src/data/components/simpleBuy/sagas.ts
```diff
@@ -36,6 +46,10 @@ import {
 import { errorHandler } from 'blockchain-wallet-v4/src/utils'
 import { find, pathOr, propEq } from 'ramda'
 
+import { FALLBACK_DELAY, getOutputFromPair } from '../swap/model'
+import { getDirection } from './utils'
+import { getQuote } from 'blockchain-wallet-v4-frontend/src/modals/SimpleBuy/EnterAmount/Checkout/validation'
```
blockdylanb, line 51: Should this be a relative file path?

packages/blockchain-wallet-v4-frontend/src/data/components/swap/model.ts
```diff
@@ -1,5 +1,7 @@
 import { CoinType, SwapOrderType } from 'core/types'
 
+export const FALLBACK_DELAY = 10_000 * 2
```
blockdylanb, line 3: fancy

schnogz: whats the point of this?

plondon: Just easier to read numbers

schnogz (approved):

PR #2782, opened 2020-11-20
fix(swap): remove 0 if no fee, only show fee tag

TheLeoB:
## Description (optional)
https://blockchain.atlassian.net/browse/FWLT-854

## Testing Steps (optional)
Swap:

Rather than show both '0' and 'Free' tag  or network fee if wallet is custodial, only show free tag. 
Screen Shot 2020-11-20 at 11 44 17 AM


schnogz (approved):
😎

PR #2788, opened 2020-11-24
Fix/swap fee

TheLeoB:
## Description (optional)
https://blockchain.atlassian.net/browse/FWLT-864

Originally, swap designs included showing outgoing transaction fees (if on chain) in transaction feed and cancel modals. This is difficult to do, since there's no simple way to get a network fee from a specific transaction, even with the hash. Workaround for this is to link to relevant block explorer for swap transactions applicable (i.e. ON_CHAIN and FROM_USERKEY).

Also removes network fee row from cancel flyout.


dsisto-bc (approved):
LGTM!

schnogz (approved):

PR #2792, opened 2020-11-25
Wdgld annoucement banner

blockdylanb:
## Description (optional)
Adds an announcement banner to the dashboard page for wdgld.

## Testing Steps (optional)
To test look at the dashboard to see if the banner is at the top, test the "Trade wdgld Now" button and the dismiss button.


packages/blockchain-wallet-v4-frontend/src/scenes/Home/Banners/NewCurrency/index.tsx
```diff
@@ -0,0 +1,126 @@
...
+import styled from 'styled-components'
+
+import { actions, selectors } from 'data'
+import { Button, Icon, Text } from 'blockchain-info-components'
+import { CoinType, WalletCurrencyType } from 'core/types'
+import { Wrapper } from '../styles'
+
+const StyledWrapper = styled(Wrapper)`
+  align-items: center !important;
+  font-family: 'Inter', -apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto,
```
schnogz, line 14: why do we need to define `font-family`?

dsisto-bc (approved):
image

packages/blockchain-wallet-v4-frontend/src/scenes/Home/Banners/NewCurrency/index.tsx
blockdylanb: When I created the `StyledWrapper` the font was a serif font. I can dig into exactly why but in order to move forward I just found the main font we used for the app and put it there.

schnogz (approved):

PR #2797, opened 2020-11-26
Hotfix(BTC):  Fix non-custodial to non-custodial send / transaction list

schnogz:
## Description (optional)
Only notify backend of pending deposit if send is from non-custodial to custodial.  This is causing users to see fake pending transactions in their feed when they sent to non-custodial.


TheLeoB (approved):
Pulled down and tested locally, all looks good.

PR #2808, opened 2020-12-07
fix(interest): show all interest accounts 

schnogz:
## Description
- Show Interest accounts in coin dropdowns so balances add up to the total
- Update button linkout to the new transaction page for Interest


milan-bc (approved):
LGTM

PR #2825, opened 2020-12-17
fix(sell): use payment to calculate max avail

TheLeoB:
## Description (optional)
Maximum sell amount validation in the checkout form was not accounting for network fees. This checks if a payment is created (will only happen from non-custodial accounts) and uses that effective balance as a max rather than account's full balance.
## Testing Steps (optional)

### DIFF
--- a/packages/blockchain-wallet-v4-frontend/src/modals/SimpleBuy/EnterAmount/Checkout/validation.tsx
+++ b/packages/blockchain-wallet-v4-frontend/src/modals/SimpleBuy/EnterAmount/Checkout/validation.tsx
@@ -76,7 +76,7 @@ export const getMaxMin = (
   orderType: SBOrderActionType,
   QUOTE: SBQuoteType | { quote: SwapQuoteType; rate: number },
   pair: SBPairType,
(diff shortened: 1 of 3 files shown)


packages/blockchain-wallet-v4-frontend/src/modals/SimpleBuy/EnterAmount/Checkout/validation.tsx
```diff
@@ -155,6 +158,7 @@ export const getMaxMinSell = (
   orderType: SBOrderActionType,
   quote: { quote: SwapQuoteType; rate: number },
   pair: SBPairType,
+  payment: undefined | PaymentValue,
```
schnogz, line 161: do we need explicitly to define `undefined`? Can it just be `payment?: PaymentValue,`?

schnogz: same question for line 79

schnogz (approved):

PR #2836, opened 2020-12-23
Fix/eth check

TheLeoB:
## Description (optional)
Sell and Swap 2.0: User would be able to proceed with non-custodial sends for ERC20 tokens even if their wallet didn't have enough eth for the tx. This disables at button the checkout screen for both with an error explaining why. 

## Testing Steps (optional)

### DIFF
--- a/packages/blockchain-wallet-v4-frontend/src/modals/SimpleBuy/EnterAmount/Checkout/template.success.tsx
+++ b/packages/blockchain-wallet-v4-frontend/src/modals/SimpleBuy/EnterAmount/Checkout/template.success.tsx
@@ -104,13 +104,16 @@ const ErrorTextContainer = styled.div`
   display: flex;
   justify-content: center;
   flex-direction: row;
+  margin-left: 40px;
+  margin-right: 40px;
 `
 const ErrorText = styled(Text)`
   display: inline-flex;
+  align-items: center;
   font-weight: 500;
   font-size: 14px;
   padding: 6px 12px;
-  border-radius: 32px;
+  border-radius: 8px;
   background-color: ${props => props.theme.red000};
   color: ${props => props.theme.red800};
   margin-bottom: 16px;
@@ -286,18 +289,18 @@ const Success: React.FC<InjectedFormProps<{}, Props> & Props> = props => {
       : amountRowNode.children[amountRowNode.children.length - 1]
     currencyNode.style.fontSize = `${fontSizeNumber * fontRatio}px`
   }
-
   const limit = Number(props.sddLimit.max) / SDD_LIMIT_FACTOR
+  // if user is attempting to send NC ERC20, ensure they have sufficient
+  // ETH balance else warn user and disable trade
   const isErc20 = props.supportedCoins[cryptoCurrency].contractAddress
   const isSufficientEthForErc20 =
     props.payment &&
-    (props.payment.coin === 'PAX' ||
-      props.payment.coin === 'USDT' ||
-      props.payment.coin === 'WDGLD') &&
-    !props.payment.isSufficientEthForErc20 &&
-    isErc20 &&
     props.swapAccount?.type === 'ACCOUNT' &&
-    props.orderType === 'SELL'
+    props.orderType === 'SELL' &&
+    isErc20 &&
+    // @ts-ignore
+    !props.payment.isSufficientEthForErc20
+
   return (
     <CustomForm onSubmit={props.handleSubmit}>
       <FlyoutWrapper style={{ paddingBottom: '0px', borderBottom: 'grey000' }}>
--- a/packages/blockchain-wallet-v4-frontend/src/modals/SimpleBuy/EnterAmount/Checkout/template.success.tsx
+++ b/packages/blockchain-wallet-v4-frontend/src/modals/SimpleBuy/EnterAmount/Checkout/template.success.tsx
@@ -288,7 +288,16 @@ const Success: React.FC<InjectedFormProps<{}, Props> & Props> = props => {
   }
 
   const limit = Number(props.sddLimit.max) / SDD_LIMIT_FACTOR
-
+  const isErc20 = props.supportedCoins[cryptoCurrency].contractAddress
+  const isSufficientEthForErc20 =
+    props.payment &&
+    (props.payment.coin === 'PAX' ||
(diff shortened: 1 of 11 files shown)


packages/blockchain-wallet-v4-frontend/src/modals/SimpleBuy/EnterAmount/Checkout/template.success.tsx
```diff
@@ -288,7 +288,20 @@ const Success: React.FC<InjectedFormProps<{}, Props> & Props> = props => {
...
+  const isErc20 =
+    props.payment &&
+    (props.payment.coin === 'PAX' ||
+      props.payment.coin === 'USDT' ||
+      props.payment.coin === 'WDGLD')
+  const isSufficientEthForErc20 =
+    props.payment &&
+    (props.payment.coin === 'PAX' ||
+      props.payment.coin === 'USDT' ||
+      props.payment.coin === 'WDGLD') &&
```
schnogz, line 300: shouldnt need to repeat this check again. I think you can just use the flag set above?

packages/blockchain-wallet-v4-frontend/src/modals/SimpleBuy/EnterAmount/Checkout/template.success.tsx
```diff
@@ -500,11 +513,32 @@ const Success: React.FC<InjectedFormProps<{}, Props> & Props> = props => {
             </ErrorText>
           </ErrorTextContainer>
         )}
-        <ActionButton {...props} />
+        <ActionButton
+          {...props}
+          isSufficientEthForErc20={isSufficientEthForErc20 || false}
```
schnogz, line 518: i dont think you need the `|| false`. It should already be a boolean or at least undefined which is falsey value

packages/blockchain-wallet-v4-frontend/src/modals/SimpleBuy/EnterAmount/Checkout/template.success.tsx
```diff
@@ -500,11 +513,32 @@ const Success: React.FC<InjectedFormProps<{}, Props> & Props> = props => {
...
+        <ErrorTextContainer>
+          <ErrorText>
+            <Icon
+              name='alert-filled'
+              color='red600'
+              style={{ marginRight: '4px' }}
+            />
+            <FormattedMessage
+              id='modals.interest.deposit.notenougheth'
+              defaultMessage='ETH is required to send {coinTicker}. You do not have enough ETH to perform a transaction.'
```
schnogz, line 534: i think we need to explicitly state that the ETH needs to be in the **non-custodial** wallet and not the custodial.

packages/blockchain-wallet-v4-frontend/src/modals/Swap/EnterAmount/Checkout/index.tsx
```diff
@@ -216,7 +216,16 @@ const Checkout: React.FC<InjectedFormProps<{}, Props> & Props> = props => {
   const balanceBelowMinimum = userMax < Number(min)
 
   const isQuoteFailed = Remote.Failure.is(props.quoteR)
-
+  const isErc20 =
+    BASE.coin === 'PAX' || BASE.coin === 'USDT' || BASE.coin === 'WDGLD'
```
schnogz, line 220: use the `contractAddress` flag in wallet-options similar to above

packages/blockchain-wallet-v4-frontend/src/modals/Swap/EnterAmount/Checkout/index.tsx
```diff
@@ -465,6 +474,17 @@ const Checkout: React.FC<InjectedFormProps<{}, Props> & Props> = props => {
             })}
           </ErrorCartridge>
         )}
+        {disableInsufficientEth && (
+          <ErrorCartridge style={{ marginTop: '16px' }}>
+            <FormattedMessage
+              id='modals.interest.deposit.notenougheth'
+              defaultMessage='ETH is required to send {coinTicker}. You do not have enough ETH to perform a transaction.'
```
schnogz, line 481: update the messaging here as well

schnogz (changes requested):

packages/blockchain-wallet-v4-frontend/src/modals/SimpleBuy/EnterAmount/Checkout/template.success.tsx
```diff
@@ -288,7 +288,20 @@ const Success: React.FC<InjectedFormProps<{}, Props> & Props> = props => {
   }
 
   const limit = Number(props.sddLimit.max) / SDD_LIMIT_FACTOR
-
+  const isErc20 =
+    props.payment &&
+    (props.payment.coin === 'PAX' ||
+      props.payment.coin === 'USDT' ||
+      props.payment.coin === 'WDGLD')
```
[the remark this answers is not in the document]
TheLeoB: Nice, thanks.

packages/blockchain-wallet-v4-frontend/src/modals/SimpleBuy/EnterAmount/Checkout/template.success.tsx
TheLeoB: The check needs to be repeated, because `isSufficientEthForErc20` doesn't exist on the BtcPaymentValue, so in the declaration, it needs to be v. explicit that it's an Eth payment value. I tried removing it and received a TS error.

packages/blockchain-wallet-v4-frontend/src/modals/SimpleBuy/EnterAmount/Checkout/template.success.tsx
TheLeoB: If i remove the `||false`, i get an error since I set the TS type to boolean so it doesn't like it. What I could do is set the type to boolean or undefined, is that better?

milan-bc: what about converting `isSufficientEthForErc20` to boolean like Boolean(isSufficientEthForErc20)?

schnogz (approved):

PR #2853, opened 2021-01-11
feat(veriff): prevent injecting more than one iframe

milan-bc:
## Description (optional)
This PR prevents rendering more than one iFrame from veriff

## Testing Steps (optional)
Try to verify gold and make sure that each time proper veriff modal is visible and usable.


dsisto-bc (approved):
LGTM. I had a similar problem in exchange

milan-bc:
@dsisto-bc yeah pretty nasty one. I red veriff docs and they do not have anything to prevent such thing :(

PR #2854, opened 2021-01-11
feat(sdd): prevent sdd flow and prevent tier 3 sdd user to add new card

milan-bc:
## Description (optional)

## Testing Steps (optional)

### DIFF
--- a/packages/blockchain-wallet-v4-frontend/src/modals/SimpleBuy/CheckoutConfirm/index.tsx
+++ b/packages/blockchain-wallet-v4-frontend/src/modals/SimpleBuy/CheckoutConfirm/index.tsx
@@ -13,7 +13,6 @@ import {
   WalletFiatType
 } from 'core/types'
 import { getFiatFromPair, getOrderType } from 'data/components/simpleBuy/model'
-import { Remote } from 'blockchain-wallet-v4/src'
 import { RootState } from 'data/rootReducer'
 import { UserDataType } from 'data/types'
 import DataError from 'components/DataError'
@@ -32,9 +31,6 @@ class CheckoutConfirm extends PureComponent<Props> {
       this.props.order.inputQuantity
     )
     this.props.sendActions.getLockRule()
-    if (!Remote.Success.is(this.props.data)) {
-      this.props.simpleBuyActions.fetchBankTransferAccounts()
-    }
   }
 
   handleSubmit = () => {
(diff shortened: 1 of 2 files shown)


packages/blockchain-wallet-v4-frontend/src/modals/SimpleBuy/CheckoutConfirm/selectors.ts
```diff
@@ -13,45 +13,20 @@ export const getData = (state: RootState) => {
...
-      isUserSddVerified: ExtractSuccess<typeof isUserSddVerifiedR>,
-      cards: ExtractSuccess<typeof cardsR>
+      withdrawLockCheck: ExtractSuccess<typeof withdrawLockCheckR>
     ) => ({
       bankAccounts,
       quote,
       sbBalances,
       userData,
-      withdrawLockCheck,
-      isSddFlow: sddEligible.eligible || userSDDTier === 3,
```
milan-bc, line 42: main issue was with this one, anyhow since we do not need it at the moment I removed all SDD related stuff.
In case that it's hard change we can revert this back and make sure that user tier do not be included in this check

packages/blockchain-wallet-v4-frontend/src/modals/SimpleBuy/CheckoutConfirm/index.tsx
```diff
@@ -33,48 +33,18 @@ class CheckoutConfirm extends PureComponent<Props> {
     )
     this.props.sendActions.getLockRule()
     if (!Remote.Success.is(this.props.data)) {
-      this.props.simpleBuyActions.fetchSDDEligible()
-      this.props.simpleBuyActions.fetchSDDVerified()
-      this.props.simpleBuyActions.fetchSBCards()
```
schnogz, line 38: why are we removing fetching cards?

packages/blockchain-wallet-v4-frontend/src/modals/SimpleBuy/CheckoutConfirm/selectors.ts
schnogz: if that is the issue, can we just set it to `false`?

milan-bc: yes it will work

packages/blockchain-wallet-v4-frontend/src/modals/SimpleBuy/CheckoutConfirm/index.tsx
milan-bc: cards were used only for sdd part since this PR removes all of that we can remove loading cards at this point

blockdylanb: Ah, I thought we were fetching cards on checkout confirm just in case any of them got blocked or voided before the user confirmed their buy. If that's not the case then you can probably remove `this.props.simpleBuyActions.fetchBankTransferAccounts()` too

milan-bc: yup I removed this one easily

schnogz (approved):

PR #2859, opened 2021-01-12
fix(bank linking): fixes issue where UI was not reflecting the proper…

blockdylanb:
## Description (optional)
When linking a bank account if there was a linking error from the api, the frontend would only show the default error instead of showing issue specific error fly-outs. This fixes that by properly passing the api error code to the template to render the proper error text.

## Testing Steps (optional)
You'll have to add a mocked version of `bankData` in `simplBuy/sagas.ts` which has `state: 'BLOCKED'` and a `error: 'BANK_TRANSFER_ACCOUNT_ALREADY_LINKED'`


schnogz (approved):
like the use of `T`. 👏

PR #2861, opened 2021-01-13
fix(payment methods): use new payment method eligibity endpoint

schnogz:
## Description
- Fix to ensure non-gold users are shown payment methods for Buy/Sell instead of the currency unsupported screen
- Removes duplicate entries of same APIs.  The two removed where not used in the codebase


blockdylanb (commented):
Looks good to me. Works locally as well. I'll let Milan take a look and approve if he doesn't have any objections.

milan-bc (approved):
Tested and code reviewed LGTM!

PR #2865, opened 2021-01-14
fix(settings): Merge linked banks and beneficiaries

blockdylanb:
## Description (optional)
Instead of showing bank wire accounts and ach accounts in separate sections we just want to show them in one section. I attempted to merge the data into one component, but figured it would be much simpler to just move the bank wire section under the linked banks section and remove the title so it looks like they're in the same list.

https://share.getcloudapp.com/kpuNq5o8

## Testing Steps (optional)
Add ACH account and bank wire account and view them in the `/settings/general` page


blockdylanb:
beneficiaries endpoint is 500ing in staging so i haven't been able to test properly yet.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "blockchain-wallet-v4",
-  "version": "4.47.12",
+  "version": "4.47.13",
```
schnogz, line 3: undo this change since it gets done automagically via `yarn release`

schnogz (changes requested):

PR #2866, opened 2021-01-15
fix(ACH): Payment methods defaults and order history

schnogz:
## Description (optional)
- Show correct bank account information on transaction list instead of "Bank Account"
- By default select last bank account used for Buy
- Fix Dark mode grey900 text


milan-bc (approved):
Looks good, let's roll

PR #2867, opened 2021-01-18
feat(swap): prevent caching of old swap amount after success swap

milan-bc:
## Description (optional)
This PR prevents caching of swap initial value, after success swap we now do clear previously entered amount

## Testing Steps (optional)

### DIFF
--- a/packages/blockchain-wallet-v4-frontend/src/data/components/swap/sagas.ts
+++ b/packages/blockchain-wallet-v4-frontend/src/data/components/swap/sagas.ts
@@ -222,6 +222,7 @@ export default ({
       }
       yield put(actions.form.stopSubmit('previewSwap'))
       yield put(actions.components.refresh.refreshClicked())
+      yield put(actions.form.destroy('swapAmount'))
       yield put(
         A.setStep({
           step: 'SUCCESSFUL_SWAP',


packages/blockchain-wallet-v4-frontend/src/data/components/swap/sagas.ts
```diff
@@ -222,6 +222,7 @@ export default ({
       }
       yield put(actions.form.stopSubmit('previewSwap'))
       yield put(actions.components.refresh.refreshClicked())
+      yield put(actions.form.destroy('swapAmount'))
```
dsisto-bc, line 225: May or may not help: I think this is automatically done when unmounting the reduxForm component. Maybe your component has `destroyOnUnmount: false`.

dsisto-bc (approved):
LGTM! But probably someone with more knowledge on the impact of wallet forms should have a look

packages/blockchain-wallet-v4-frontend/src/data/components/swap/sagas.ts
schnogz: i believe it does, @milan-bc can you check that. removing that would be the correct way to fix this

schnogz (changes requested):

packages/blockchain-wallet-v4-frontend/src/data/components/swap/sagas.ts
milan-bc: we do have it `destroyOnUnmount: false` but what I notice then if you go forward and want to return back to enter amount page we lost entered amount in that case, that is why I added destroy after we do create order

milan-bc: we have preview swap where user can decide to go back, UX thingy we have to keep form value on unmount in such case

PR #2869, opened 2021-01-19
fix(ach): style changes for simple buy modal

blockdylanb:
## Description (optional)
Design changes to make arrow icons bigger on simple buy flow, remove the "x" from the top right, change FIAT/Crypto toggle icon, and make max buy/sell button not display actual ammounts

## Testing Steps (optional)
Open simple buy flow and it should look more like this: 
image


packages/blockchain-wallet-v4-frontend/src/modals/SimpleBuy/EnterAmount/Checkout/Payment/index.tsx
```diff
@@ -18,6 +18,7 @@ import { Props } from '../template.success'
 const RightArrowIcon = styled(Icon)<{
   disabled?: boolean
 }>`
+  transform: rotate(180deg);
```
blockdylanb, line 21: I tried all the different "right" arrows we had, but none of them were the larger skinnier ones so I just used the "arrow-back" icon and rotated it to get the desired style.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "blockchain-wallet-v4",
-  "version": "4.47.12",
+  "version": "4.47.13",
```
schnogz, line 3: undo

schnogz (changes requested):

package.json
blockdylanb: done

schnogz (approved):

PR #2878, opened 2021-01-22
fix(tx list): always have a default case so exotic order types dont blow up app

schnogz:
Screen Shot 2021-01-21 at 7 13 05 PM


blockdylanb (approved):
LGTM. Could we add the default check to the linter so it'll scream at us if someone ever forgets?

milan-bc (approved):

PR #2879, opened 2021-01-22
fix(simple buy): fixes error on first-time user credit card form validation

blockdylanb:
## Description (optional)
When a user first signs up and proceeds to go through the simple buy flow, we get the payment methods eligibility from the api _before_ the user has entered any address/kyc info. So it shows them as ineligible unless they hard refresh and log in again. This change makes the payments method fetch each time the CC form is loaded so we always have the latest eligibility. 

## Testing Steps (optional)
1. Sign up for a new wallet (DO NOT HARD REFRESH)
2. Go through simple buy flow
3. CC form should no longer show the `Card type not supported` input error when you put in a CC#


milan-bc (approved):

schnogz:
What about existing users? Does everything still work?

khokho1 (approved):

PR #2884, opened 2021-01-25
fix(tx-feed): pax to usd digital display

TheLeoB:
## Description (optional)
Fixed 985, where swap/buy/sell transactions displayed PAX in the feed rather than usd-d. 

## Testing Steps (optional)

### DIFF
--- a/packages/blockchain-wallet-v4-frontend/src/scenes/Transactions/CustodialTx/model.tsx
+++ b/packages/blockchain-wallet-v4-frontend/src/scenes/Transactions/CustodialTx/model.tsx
@@ -152,18 +152,21 @@ export const Origin = (props: Props) => {
     case 'REFUNDED':
     case 'DEPOSIT':
       return props.tx.amount.symbol in CoinTypeEnum ? (
-        <>{props.coinTicker} Wallet</>
+        <>{props.supportedCoins[props.coin].displayName} Wallet</>
       ) : (
         <>Bank Account</>
       )
     case 'SELL':
       return props.tx.extraAttributes?.direction === 'FROM_USERKEY' ? (
-        <>{props.tx.amount.symbol} Wallet</>
+        <>{props.supportedCoins[props.tx.amount.symbol].displayName} Wallet</>
       ) : (
-        <>{props.tx.amount.symbol} Trading Wallet</>
+        <>
+          {props.supportedCoins[props.tx.amount.symbol].displayName} Trading
+          Wallet
+        </>
       )
     case 'WITHDRAWAL':
-      return <>{props.coinTicker} Wallet</>
+      return <>{props.supportedCoins[props.coin].displayName} Wallet</>
     default:
       return <></>
   }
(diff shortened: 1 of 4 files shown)


packages/blockchain-wallet-v4-frontend/src/scenes/Transactions/CustodialTx/model.tsx
```diff
@@ -152,18 +152,21 @@ export const Origin = (props: Props) => {
...
       return props.tx.amount.symbol in CoinTypeEnum ? (
-        <>{props.coinTicker} Wallet</>
+        <>{props.supportedCoins[props.coin].displayName} Wallet</>
       ) : (
         <>Bank Account</>
       )
     case 'SELL':
       return props.tx.extraAttributes?.direction === 'FROM_USERKEY' ? (
-        <>{props.tx.amount.symbol} Wallet</>
+        <>{props.supportedCoins[props.tx.amount.symbol].displayName} Wallet</>
```
blockdylanb, line 161: I guess this should be fine if typscript didn't complain about possible undefined keys here, but doing this many dot notations is a little scary. Maybe useful to use Ramda `path` or `defaultTo` just in case one of those is undefined.

packages/blockchain-wallet-v4-frontend/src/scenes/Transactions/SwapOrderTx/model.tsx
```diff
@@ -17,10 +17,15 @@ export const getDestination = (props: Props) => {
...
     case 'ON_CHAIN':
-      return getOutput(props.order) + ' Wallet'
+      return (
+        props.supportedCoins[getOutput(props.order)].displayName + ' Wallet'
+      )
     case 'FROM_USERKEY':
     case 'INTERNAL':
-      return getOutput(props.order) + ' Trading Wallet'
+      return (
+        props.supportedCoins[getOutput(props.order)].displayName +
```
blockdylanb, line 26: Maybe try pulling `props.supportedCoins[getOutput(props.order)].displayName` into a variable and use template string interpolation instead of `+`

packages/blockchain-wallet-v4-frontend/src/scenes/Transactions/SwapOrderTx/model.tsx
```diff
@@ -9,6 +10,15 @@ import {
 } from '../components'
 import { Text } from 'blockchain-info-components'
 
+const getOutputCoinDisplayName = (props: Props) => {
+  return path(
+    [`${getOutput(props.order)}`, 'displayName'],
+    props.supportedCoins
+  )
+}
```
milan-bc, line 18: this is same function as `getSymbolDisplayName` might be good to move it to some more common place and just reuse

schnogz: yeah i would put the shared function in `packages/blockchain-wallet-v4-frontend/src/scenes/Transactions/model.tsx` and then import into the 3 templates that need it

packages/blockchain-wallet-v4-frontend/src/scenes/Transactions/CustodialTx/model.tsx
```diff
@@ -16,7 +17,16 @@ const Icon = styled(BCIcon)`
   size: 18px;
   font-weight: 600;
 `
+const getSymbolDisplayName = (props: Props) => {
+  return path(
+    [`${props.tx.amount.symbol}`, 'displayName'],
```
schnogz, line 22: i dont think you need to do the string template. just the following should work
```
path([props.tx.amount.symbol, 'displayName'],
````

File: packages/blockchain-wallet-v4-frontend/src/scenes/Transactions/CustodialTx/model.tsx [hunk 5]
```diff
@@ -16,7 +17,16 @@ const Icon = styled(BCIcon)`
...
+    props.supportedCoins
+  )
+}
 
+const getCoinDisplayName = (props: Props) => {
+  return path([`${props.coin}`, 'displayName'], props.supportedCoins)
```
schnogz, line 28: same here and probably in other spots

blockdylanb (approved):

PR #2886, opened 2021-01-25
fix(simple buy): fixes issue where funds default payment method has no available funds

blockdylanb:
## Description (optional)
If you have fiat funds, and buy crypto with 100% of your fiat wallet, the next time you try to buy crypto, it defaults to previous fiat wallet, but says I don't have enough funds, but you can't change the payment method because there's a form error.

## Testing Steps (optional)
Add fiat funds to your account, buy crypto with 100% of those funds. Try to buy more crypto again.


blockdylanb:
was able to test this today, and it successfully removes the default fiat payment method if there are no available funds.

schnogz (approved):

PR #2887, opened 2021-01-25
fix(credit cards): remove click event from cards in general settings

blockdylanb:
## Description (optional)
There's a click event on all cards in general settings that just opens the simple buy flyout. This should not be there.

## Testing Steps (optional)
Add CC to an account or use existing CC account, go to General settings, click on the save CC (not the remove button) and it should now do nothing.


milan-bc (approved):

Johnny-hub (approved):

dsisto-bc (approved):
LGTM!

PR #2889, opened 2021-01-26
fix(swap): disable swap if avail balance is less than min swap

TheLeoB:
## Description (optional)
Disable swap if user's wallet balance < swap minimum. Was originally comparing wallet balance in minor units to minimum swap (major units), so the check was always returning false. This change fixes the check to use the correctly converted wallet balance. 

## Testing Steps (optional)


schnogz (approved):
🎉

PR #2891, opened 2021-01-27
fix(cards): add new button in general settings should open the flyout straight to the cc form

blockdylanb:
## Description (optional)
Need ability to add a new credit card from the user's general settings. Currently you're just put into the simple buy flow which tries to make you place an order which is not what the user expects.
This change puts the user into the simple buy flow but starts them at the CC form step, then short circuits them out of the simple buy modal flow once the card has been successfully added.

## Testing Steps (optional)
- Login
- Go to general settings
- Click "Add Card"
- Add a test card
- You should only go through the card adding steps and not the normal simple buy flow


milan-bc (approved):
LGTM

yichok:
No

PR #2897, opened 2021-01-29
feat(borrow): disable new loan creation while feature is being sunset

schnogz:
## Description
Borrow is being decommissioned and no new loans are allowed to be created. The feature will be fully removed once all open loans are closed.


packages/blockchain-wallet-v4-frontend/src/scenes/Borrow/InitBorrowForm/index.tsx
```diff
@@ -201,7 +178,7 @@ type LinkStatePropsType = {
   }
 }
 
-type Props = OwnProps & ConnectedProps<typeof connector>
+type Props = ConnectedProps<typeof connector>
```
milan-bc, line 181: 👏

milan-bc (approved):
LGTM

PR #2903, opened 2021-02-01
fix(sdd): once when we know that user is verified we set this flag in state

milan-bc:
## Description (optional)
This PR fix a multiple calls to /nabu-gateway/sdd/verified in case that user is already verified we do not need to again call BE anymore since user can't lost verification

## Testing Steps (optional)
- create a fresh wallet
- go over SDD
- after you enter dob/address make sure that in network tab we do not call more calls to /nabu-gateway/sdd/verified after BE return that user is verified

### DIFF
--- a/packages/blockchain-wallet-v4-frontend/src/data/components/identityVerification/sagas.ts
+++ b/packages/blockchain-wallet-v4-frontend/src/data/components/identityVerification/sagas.ts
@@ -418,7 +418,12 @@ export default ({ api, coreSagas, networks }) => {
             break
           }
           sddVerified = yield call(api.fetchSDDVerified)
-          if (sddVerified?.taskComplete) break
+          if (sddVerified?.taskComplete) {
+            yield put(
+              actions.components.simpleBuy.fetchSDDVerifiedSuccess(sddVerified)
+            )
+            break
+          }
           yield delay(POLL_SDD_DELAY)
         }
 
--- a/packages/blockchain-wallet-v4-frontend/src/data/components/simpleBuy/sagas.ts
+++ b/packages/blockchain-wallet-v4-frontend/src/data/components/simpleBuy/sagas.ts
@@ -752,9 +752,15 @@ export default ({
   }
   const fetchSDDVerified = function * () {
     try {
-      yield put(A.fetchSDDVerifiedLoading())
-      const sddEligible = yield call(api.fetchSDDVerified)
-      yield put(A.fetchSDDVerifiedSuccess(sddEligible))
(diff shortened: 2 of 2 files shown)


packages/blockchain-wallet-v4-frontend/src/data/components/identityVerification/sagas.ts
```diff
@@ -418,7 +418,12 @@ export default ({ api, coreSagas, networks }) => {
             break
           }
           sddVerified = yield call(api.fetchSDDVerified)
-          if (sddVerified?.taskComplete) break
+          if (sddVerified?.taskComplete) {
+            yield put(
+              actions.components.simpleBuy.fetchSDDVerifiedSuccess(sddVerified)
+            )
```
milan-bc, line 424: we got response that user is verified, flag this in SimpleBuy flow (maybe we should move this to more common place)

packages/blockchain-wallet-v4-frontend/src/data/components/simpleBuy/sagas.ts
```diff
@@ -752,9 +752,15 @@ export default ({
   }
   const fetchSDDVerified = function * () {
     try {
-      yield put(A.fetchSDDVerifiedLoading())
-      const sddEligible = yield call(api.fetchSDDVerified)
-      yield put(A.fetchSDDVerifiedSuccess(sddEligible))
+      const isSddVerified = S.getSddVerified(yield select()).getOrElse({
+        verified: false
+      })
```
milan-bc, line 757: I added fall back and then we do call API otherwise if user is already verified we do not need to call API endpoint.

packages/blockchain-wallet-v4-frontend/src/data/components/identityVerification/sagas.ts
```diff
@@ -418,7 +418,12 @@ export default ({ api, coreSagas, networks }) => {
             break
```
schnogz, line 418: do you need to add that logic here as well?

schnogz (changes requested):
per my question above

packages/blockchain-wallet-v4-frontend/src/data/components/identityVerification/sagas.ts
milan-bc: in general we can if we want to prevent additional calls if saga trigger 10 times, I'm not sure is it possible to happen that we trigger 10 times and BE is still not really processed this we mark it and then when user enter other page we do call to BE it user is verified now. I think if we add same logic we might lost this edge case.

schnogz (approved):

PR #2906, opened 2021-02-03
fix(sell): sell for should also stop submit

TheLeoB:
## Description (optional)
Original Issue: If /trades returned an error for sell, the flow would return to Checkout Screen but form would stay in loading state and error message would never show.

Caused by a start submit for the simpleBuyCheckout form on line 391 that was never stopped. Removed the 'return' from line 397 so that saga would go onto 401 and stop the submit.

## Testing Steps (optional)


blockdylanb (approved):

yichok (commented):
Yy

PR #2924, opened 2021-02-16
Fixing `.prettierrc` config

pedroapfilho:
This is just a rule to make `vscode` and other text editors to comply with `prettier`. It seems that they've created a new rule for `jsx`, and now you have to explicitly add `jsxSingleQuote` as `true`, in order to get single quotes on `tsx` or `jsx`.


schnogz (approved):
Nice one!

PR #2955, opened 2021-03-01
Fix/segwit swap balance

TheLeoB:
## Description (optional)
1. Update imports on some files, I think were added as part of merge from refactors branch. Tests were failing because of this.
2. Some gnarly logic to change how we determine accounts balance for swap. Structure of accounts have changed to include `derivations` key, and we need to get both xPub for each account, and then add the two balances together for total. 
## Testing Steps (optional)

### DIFF
--- a/packages/blockchain-wallet-v4-frontend/src/data/coins/selectors/coins/btc.tsx
+++ b/packages/blockchain-wallet-v4-frontend/src/data/coins/selectors/coins/btc.tsx
@@ -1,5 +1,5 @@
+import { add, lift, prop, propEq } from 'ramda'
 import { FormattedMessage } from 'react-intl'
-import { lift, prop, propEq } from 'ramda'
 import React from 'react'
 
 import { ADDRESS_TYPES } from 'blockchain-wallet-v4/src/redux/payment/btc/utils'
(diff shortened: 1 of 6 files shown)


packages/blockchain-wallet-v4-frontend/src/data/coins/selectors/coins/btc.tsx
```diff
@@ -37,42 +37,52 @@ export const getAccounts = createDeepEqualSelector(
...
-            balance: prop('final_balance', prop(prop('xpub', acc), btcData)),
-            baseCoin: coin,
-            coin,
-            label: prop('label', acc) || prop('xpub', acc),
-            type: ADDRESS_TYPES.ACCOUNT
-          }))
-          .filter(propEq('archived', false)))
+        // each account has a derivations object with legacy address and segwit address
+        // need to extract each xpub for balance
+        const xpubArray = acc =>
```
TheLeoB, line 44: gnarly logic begins here. If there's a simpler way to do this, I'd love it.

schnogz: there might be a generic selector for doing this. if not we should consider making one. looking now 👀

schnogz: actually this is fine the way it is

packages/blockchain-wallet-v4-frontend/src/data/coins/selectors/coins/btc.tsx
```diff
@@ -37,42 +37,52 @@ export const getAccounts = createDeepEqualSelector(
...
-            archived: prop('archived', acc),
-            // @ts-ignore
-            balance: prop('final_balance', prop(prop('xpub', acc), btcData)),
-            baseCoin: coin,
-            coin,
-            label: prop('label', acc) || prop('xpub', acc),
-            type: ADDRESS_TYPES.ACCOUNT
-          }))
-          .filter(propEq('archived', false)))
+        // each account has a derivations object with legacy address and segwit address
```
schnogz, line 42: change comment ending to "with legacy and segwit xpubs"

schnogz (changes requested):

schnogz (approved):

PR #2967, opened 2021-03-05
Fix/ach issues

TheLeoB:
## Description (optional)
@blockdylanb 

1. Adds error fallback if deposit fails: sends user back to Enter Amount step with error message like this:
Screen Shot 2021-03-04 at 9 20 25 PM

2. Destroys all values in the brokerageTx form when 'deposit' button is clicked from the transaction feed. This is the best way I found so far to clear form values/errors. If I did it on the `handleClose` or when canceling an order, there would be a brief moment where values would show as NaN while the flyout transitioned from open to close. Not sure if this is the best way, let me know.

3. Adds currency symbol and fiat cash wallet to order success.
Screen Shot 2021-03-04 at 9 38 20 PM

## Testing Steps (optional)

### DIFF
--- a/packages/blockchain-wallet-v4-frontend/src/data/components/brokerage/sagas.ts
+++ b/packages/blockchain-wallet-v4-frontend/src/data/components/brokerage/sagas.ts
@@ -193,6 +193,7 @@ export default ({
           'BANK_DEPOSIT_MODAL'
         )
       )
+      yield put(actions.form.destroy('brokerageTx'))
       yield put(
         actions.components.brokerage.setDWStep({
           dwStep: BankDWStepType.ENTER_AMOUNT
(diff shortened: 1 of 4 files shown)


packages/blockchain-wallet-v4-frontend/src/data/components/brokerage/sagas.ts
```diff
@@ -193,6 +193,7 @@ export default ({
           'BANK_DEPOSIT_MODAL'
         )
       )
+      yield put(actions.form.destroy('brokerageTx'))
```
TheLeoB, line 196: Destroys form values before enter amount step shows.

milan-bc (approved):

schnogz (approved):

PR #2968, opened 2021-03-05
fix(brokerage): adds a fiat eligibility check when opening up the dep…

blockdylanb:
## Description (optional)
A user is eligible to make deposits and withdrawals for one fiat. We need to check their eligibility when they open up the deposit or withdrawal modal and give them an informative error screen if they're not eligible.

## Testing Steps (optional)
Open one of the fiat wallets your account is not eligible for, and click deposit or withdraw buttons in the upper right hand corner. You should seen and error screen.

### DIFF
--- a/packages/blockchain-wallet-v4-frontend/src/components/Brokerage/index.tsx
+++ b/packages/blockchain-wallet-v4-frontend/src/components/Brokerage/index.tsx
@@ -13,14 +13,18 @@ const AddNewButton = styled(Button)`
   border-color: ${props => props.theme.grey100};
 `
 
+const StyledText = styled(Text)`
+  width: 300px;
+`
+
 const BROKERAGE_INELIGIBLE = 'BROKERAGE_INELIGIBLE'
 const IneligibleErrorMessage = () => (
-  <Text size='16px' weight={400}>
+  <StyledText size='16px' weight={400}>
     <FormattedMessage
       id='modals.brokerage.ineligible_error'
       defaultMessage='You are not eligible to make deposits and withdrawals with this currency.'
     />
-  </Text>
+  </StyledText>
 )
 
 export { AddNewButton, BROKERAGE_INELIGIBLE, IneligibleErrorMessage }
--- a/packages/blockchain-wallet-v4-frontend/src/components/Brokerage/index.tsx
+++ b/packages/blockchain-wallet-v4-frontend/src/components/Brokerage/index.tsx
@@ -1,4 +1,6 @@
-import { Button } from 'blockchain-info-components'
+import { Button, Text } from 'blockchain-info-components'
+import { FormattedMessage } from 'react-intl'
+import React from 'react'
 import styled from 'styled-components'
 
 const AddNewButton = styled(Button)`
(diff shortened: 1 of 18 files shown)


packages/blockchain-wallet-v4-frontend/src/data/components/brokerage/sagas.ts
```diff
@@ -177,7 +177,31 @@ export default ({
...
   }
 
-  const handleDepositFiatClick = function * () {
+  const handleDepositFiatClick = function * ({
+    payload
+  }: ReturnType<typeof A.handleDepositFiatClick>) {
+    yield put(
+      actions.components.brokerage.showModal(
+        BrokerageModalOriginType.DEPOSIT_BUTTON,
+        'BANK_DEPOSIT_MODAL'
```
milan-bc, line 186: We use this string on few places, I was thinking to move it to constant and use it like that all around.

packages/blockchain-wallet-v4-frontend/src/data/components/withdraw/reducers.ts
```diff
@@ -1,12 +1,12 @@
...
-import { WithdrawActionTypes, WithdrawState } from './types'
+import { WithdrawActionTypes, WithdrawState, WithdrawStepEnum } from './types'
 import Remote from 'blockchain-wallet-v4/src/remote/remote'
 
 const INITIAL_STATE: WithdrawState = {
   amount: undefined,
   beneficiary: undefined,
   fiatCurrency: 'EUR',
-  step: 'ENTER_AMOUNT',
+  step: WithdrawStepEnum.ENTER_AMOUNT,
```
milan-bc, line 9: 👏

milan-bc (commented):
I left one minor thing, it could be addressed as part of this PR, rest looks neat!

packages/blockchain-wallet-v4-frontend/src/components/Brokerage/index.tsx
```diff
@@ -11,4 +13,18 @@ const AddNewButton = styled(Button)`
...
+const IneligibleErrorMessage = () => (
+  <StyledText size='16px' weight={400}>
+    <FormattedMessage
+      id='modals.brokerage.ineligible_error'
+      defaultMessage='You are not eligible to make deposits and withdrawals with this currency.'
+    />
+  </StyledText>
+)
+
+export { AddNewButton, BROKERAGE_INELIGIBLE, IneligibleErrorMessage }
```
schnogz, line 30: lets move this file into a new components folder within `modals/brokerage`.

schnogz (changes requested):
nice job on the `WithdrawStepEnum`

packages/blockchain-wallet-v4-frontend/src/components/Brokerage/index.tsx
schnogz: `AddNewButton` margin above can just be `margin: 40px auto`

packages/blockchain-wallet-v4-frontend/src/data/components/brokerage/sagas.ts
blockdylanb: We should make a separate tech debt ticket for this because 'BANK_DEPOSIT_MODAL' is part of the `ModalNamesType` which has about ~45 different modal names in it. To turn that into an enum and then make the changes everywhere in the app would be a pretty big task!

schnogz (approved):

PR #2974, opened 2021-03-08
fix(codebase): removing warnings

pedroapfilho:
Fixed warnings for `componentWillUpdate` and `caniuse`

### DIFF
--- a/packages/blockchain-wallet-v4-frontend/src/modals/Lockbox/LockboxSetup/index.js
+++ b/packages/blockchain-wallet-v4-frontend/src/modals/Lockbox/LockboxSetup/index.js
@@ -19,7 +19,7 @@ import SetupTypeStep from './SetupTypeStep'
 import SoftwareDownloadStep from './SoftwareDownloadStep'
 
 class LockboxSetupContainer extends React.PureComponent {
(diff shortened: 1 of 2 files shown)


packages/blockchain-wallet-v4-frontend/src/modals/Lockbox/LockboxSetup/index.js
```diff
@@ -19,7 +19,7 @@ import SetupTypeStep from './SetupTypeStep'
 import SoftwareDownloadStep from './SoftwareDownloadStep'
 
 class LockboxSetupContainer extends React.PureComponent {
-  componentWillMount () {
+  UNSAFE_componentWillMount () {
```
milan-bc, line 22: maybe to replace it with `getDerivedStateFromProps` ?

pedroapfilho: Well, the correct would be to refactor this component to a `FC`, and get rid of these lifecycles. I think that this is good for now, as we will probably refactor it in the future.

blockdylanb (approved):

PR #2975, opened 2021-03-08
fix(withdraw): removing height so the link can be clickable

pedroapfilho:
What was happening is that one div was overlapping the other, because of the defined height. As the background of the overlapping div was transparent, we were able to see through it, but not click on the link. I had the option to change the `z-index`, but I preferred to just remove the height (unnecessary and probably just copied and pasted), to make it work.


blockdylanb (approved):

blockdylanb:
Possible to add a quick screenshot for context?

PR #2980, opened 2021-03-09
fix(mobile auth): hide email from recovery form during metadata restore

schnogz:
## Description
- When user creates wallet with new clients, recovery is stored encrypted in metadata.  When they go to recover from mnemonic seed, we check kvstore for existing metadata entries.  If we have this, we need to hide the email from the recover form.

## Testing Steps
- Create wallet with recovery stored in metadata, recover via web


milan-bc (approved):
It looks ok code wise, but I'm not quite sure how to test it properly.

PR #2982, opened 2021-03-09
feat(ach): load fiatCurrency based on page where user did clicked

milan-bc:
## Description (optional)
This PR load proper fiat currency for deposit/withdrawal click

## Testing Steps (optional)
- login with wallet which have USD account, try to use deposit/withdrawal they should work as before, the only difference is that we now load fiat currency it from store it's not hardcoded anymore

### DIFF
--- a/packages/blockchain-wallet-v4-frontend/src/data/components/brokerage/reducers.ts
+++ b/packages/blockchain-wallet-v4-frontend/src/data/components/brokerage/reducers.ts
@@ -15,7 +15,8 @@ const INITIAL_STATE: BrokerageState = {
   account: undefined,
   redirectBackToStep: false,
   addNew: false, // TODO: Put this stuff in redux-form
-  bankStatus: Remote.NotAsked
+  bankStatus: Remote.NotAsked,
+  fiatCurrency: undefined
 }
 
 export function brokerageReducer (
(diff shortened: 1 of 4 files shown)


packages/blockchain-wallet-v4-frontend/src/modals/Brokerage/Banks/Deposit/index.tsx
```diff
@@ -125,7 +125,7 @@ class Deposit extends PureComponent<Props> {
 
 const mapStateToProps = (state: RootState) => ({
   step: selectors.components.brokerage.getDWStep(state),
-  fiatCurrency: 'USD' as FiatType // TODO: Unhardcode this
+  fiatCurrency: selectors.components.brokerage.getFiatCurrency(state)
```
blockdylanb, line 128: 👏

packages/blockchain-wallet-v4-frontend/src/data/components/brokerage/reducers.ts
```diff
@@ -15,7 +15,8 @@ const INITIAL_STATE: BrokerageState = {
   account: undefined,
   redirectBackToStep: false,
   addNew: false, // TODO: Put this stuff in redux-form
-  bankStatus: Remote.NotAsked
+  bankStatus: Remote.NotAsked,
+  fiatCurrency: undefined
```
blockdylanb, line 19: Let's make a not that when we pull out cards from simple buy into brokerage we can reuse this since there is a `fiatCurrency` defined in the simplebuy state as well.

blockdylanb (approved):

PR #2984, opened 2021-03-09
fix(ach withdrawals): form should check for existence of both beneficiary and ach bank

blockdylanb:
## Description (optional)
Withdrawal enter amount screen needs to be disabled if there is no wire account and no ach accout.

## Testing Steps (optional)
try to withdraw with only a ach account and no wire accounts on your user account, you should still see the "Next" button as enabled.


TheLeoB (approved):
✅

pedroapfilho (approved):

PR #2989, opened 2021-03-10
fix(interest): wrong currency due to loading state on getOrElse

pedroapfilho:
Screen Shot 2021-03-10 at 2 05 29 PM

This was an interesting one. Due to a race condition, and the switch between Loading and Success states, the `walletCurrency` was defaulting to `GBP`, meaning that it would make the call (for the first load) as `GBP`.

It's fixed now, as we don't change the `walletCurrency` depending on the other calls.


plondon (approved):
LGTM

PR #3001, opened 2021-03-15
feat(sdd): show proper goal if user came from exchange linking

milan-bc:
## Description (optional)
This PR helps to run goals at proper time when wallet is all set

## Testing Steps (optional)
Create an exchange account and then link it with wallet, create new wallet and flyout with wallet linking should be visible


packages/blockchain-wallet-v4-frontend/src/modals/SimpleBuy/index.tsx
```diff
@@ -49,6 +49,9 @@ class SimpleBuy extends PureComponent<Props, State> {
     /* eslint-disable */
     this.setState({ show: true })
     /* eslint-enable */
+    if (this.props.isFirstLogin) {
```
schnogz, line 52: add comment on why we need this

schnogz (approved):

PR #3003, opened 2021-03-15
Fix/bch transactions

TheLeoB:
## Description (optional)

Fixes FWLT-1084.

In the tx feed, we first only fetch the first 10 transactions from multiaddr for btc/bch. If user then goes to filter by tx type (i.e. sent), if a sent transaction is in the first 10 transactions originally fetched, the feed is empty. A new param was added to the multiaddr call to get those type directly from the backend, rather than us doing some login on the frontend 

## Testing Steps (optional)

### DIFF
--- a/packages/blockchain-wallet-v4-frontend/src/data/components/bchTransactions/sagas.js
+++ b/packages/blockchain-wallet-v4-frontend/src/data/components/bchTransactions/sagas.js
@@ -49,6 +49,28 @@ export default () => {
             ? ''
             : payload.xpub || payload.address
           yield put(actions.core.data.bch.fetchTransactions(onlyShow, true))
+          break
+        case 'status':
+          const filter = payload => {
+            switch (payload) {
+              case 'sent':
+                return 1
+              case 'received':
+                return 2
+              case 'transferred':
+                return 3
+              default:
+                
+            }
+          }
+          yield put(
+            actions.core.data.bch.fetchTransactions(
+              onlyShow,
+              true,
+              filter(payload)
+            )
+          )
+          break
       }
     } catch (e) {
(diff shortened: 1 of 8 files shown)


packages/blockchain-wallet-v4-frontend/src/data/components/bchTransactions/sagas.js
```diff
@@ -49,6 +49,28 @@ export default () => {
...
+          const filter = payload => {
+            switch (payload) {
+              case 'sent':
+                return 1
+              case 'received':
+                return 2
+              case 'transferred':
+                return 3
+              default:
+
```
milan-bc, line 63: is this intentional that default do not return anything?

packages/blockchain-wallet-v4-frontend/src/data/components/btcTransactions/sagas.js
```diff
@@ -49,6 +49,28 @@ export default () => {
...
+          const filter = payload => {
+            switch (payload) {
+              case 'sent':
+                return 1
+              case 'received':
+                return 2
+              case 'transferred':
+                return 3
+              default:
+
```
milan-bc, line 63: same as above

packages/blockchain-wallet-v4-frontend/src/data/components/bchTransactions/sagas.js
TheLeoB: Yep, wanted it to just return undefined if it wasn't the 3 options. Default return is always needed for switch statements, right? I added a break as the default case.

schnogz (approved):

PR #3010, opened 2021-03-17
feat(linting): add constellation

pedroapfilho:


milan-bc:
some tests are falling but guess it's fine

milan-bc (approved):
🙏

pedroapfilho:
Ah, my bad @milan-bc I should've updated the snapshots. As the files were changed, the generated classes were changed, and the snapshots started to fail. I'll add a testing utility to parse these classes and check the actual css.

PR #3015, opened 2021-03-19
feat(datadog): added support with new header token for datadog

milan-bc:
## Description (optional)
This is support for datadog logging 

## Testing Steps (optional)
Run wallet make sure that all API calls contains in request header payload X-Request-ID


dsisto-bc (approved):
LGTM!

PR #3032, opened 2021-03-25
fix(register): dont show upgrade prompt at wallet creation

TheLeoB:
## Description (optional)
This fixes the issue where wallet creating would hang after submitting the form. We were trying to show the upgrade modal, now it will only show if it's not the user's first login. 

## Testing Steps (optional)


schnogz:
@plondon can you review?

pedroapfilho (approved):

PR #3039, opened 2021-03-29
fix(buy): fixed rendering of buy flow when bank do not have all expected details

milan-bc:
## Description (optional)
This PR resolve problem with rendering enter amount step of buy/sell flow

## Testing Steps (optional)
Login to wallet try to buy some crypto, make sure that you can proceed to enter amount screen

### DIFF
--- a/packages/blockchain-wallet-v4-frontend/src/modals/SimpleBuy/EnterAmount/Checkout/Payment/model.tsx
+++ b/packages/blockchain-wallet-v4-frontend/src/modals/SimpleBuy/EnterAmount/Checkout/Payment/model.tsx
@@ -112,9 +112,8 @@ export const renderBank = (
   <>
     <DisplayValue>{renderBankText(value)}</DisplayValue>
     <DisplayTitle>
-      {`${value.details?.bankAccountType.toLowerCase()} account ${
-        value.details?.accountNumber
-      }`}
+      {`${value.details?.bankAccountType?.toLowerCase() || ''} account ${value
+        .details?.accountNumber || ''}`}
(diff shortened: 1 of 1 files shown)


packages/blockchain-wallet-v4-frontend/src/modals/SimpleBuy/EnterAmount/Checkout/Payment/model.tsx
```diff
@@ -112,9 +112,8 @@ export const renderBank = (
   <>
     <DisplayValue>{renderBankText(value)}</DisplayValue>
     <DisplayTitle>
-      {`${value.details?.bankAccountType.toLowerCase()} account ${
-        value.details?.accountNumber
-      }`}
+      {`${value.details?.bankAccountType?.toLowerCase() || ''} account ${value
+        .details?.accountNumber || ''}`}
```
milan-bc, line 116: I put fall back to empty so we have it like. I do expect with real data to have all of those set properly anyhow this will prevent worse case scenario.  
image

schnogz (approved):

packages/blockchain-wallet-v4-frontend/src/modals/SimpleBuy/EnterAmount/Checkout/Payment/model.tsx
Fhiltyphil88: > I put fall back to empty so we have it like. I do expect with real data to have all of those set properly anyhow this will prevent worse case scenario.  
> 
> image
> 
>

PR #3041, opened 2021-03-30
feat(buy): added fallback for places where we do render bank items

milan-bc:
## Description (optional)
This PR fix all places where we do use bank object for rendering

## Testing Steps (optional)
Login to wallet and then go over sell/buy flow, make sure app is not crashing


packages/blockchain-wallet-v4-frontend/src/modals/SimpleBuy/LinkedPaymentAccounts/Accounts/Bank/index.tsx
```diff
@@ -29,7 +29,7 @@ type Props = {
   value: SBPaymentMethodType
 }
 
-const Bank: React.FC<Props> = ({ icon, onClick, text, value }) => (
+const Bank = ({ icon, onClick, text, value }: Props) => (
```
blockdylanb, line 32: 👏🏻

blockdylanb (approved):

PR #3051, opened 2021-03-31
fix/responsivity homepage chart

pedroapfilho:
Because of this problem: https://github.com/airbnb/visx/issues/1014

We can't make it 100% yet. As soon as I fix it on visx, we can do it here.


packages/blockchain-wallet-v4-frontend/src/scenes/Home/PriceChart/Chart/index.tsx
```diff
@@ -20,7 +20,7 @@ const ChartContainer = (props: Props) => {
 
   return props.data.cata({
     Success: value => (
-      <Success currency={props.currency} coin={value.coin} data={value.data} />
+      <Chart currency={props.currency} coin={value.coin} data={value.data} />
```
schnogz, line 23: 👌

schnogz (approved):

PR #3084, opened 2021-04-12
fix(animations): adding animations with framer motion

pedroapfilho:
https://user-images.githubusercontent.com/13142568/114462471-78085880-9bb9-11eb-8d67-e685532a7150.mov

https://user-images.githubusercontent.com/13142568/114462509-82c2ed80-9bb9-11eb-9121-2feb6681637d.mov

One thing that I didn't remove was the internal flyout animation library, because it would need a major refactor in order to work. With `framer-motion`, we would need to pass props to the children directly, and it would take more time than necessary. 

This PR was created to remove a deprecated lib and fix some edge cases, that after some tests, were fixed.


packages/blockchain-wallet-v4-frontend/src/components/Alerts/template.tsx
```diff
@@ -1,14 +1,12 @@
 import React from 'react'
-import ReactCSSTransitionGroup from 'react-addons-css-transition-group'
+import { AnimatePresence,motion } from 'framer-motion'
```
milan-bc, line 2: space is missing somehow should be `{ AnimatePresence, motion } `

milan-bc:
I'm seeing some slowness when opening a flyout modal (modal is flying in on slow way we might speed it up a bit maybe?), also see such error in console. 
image

packages/blockchain-info-components/index.d.ts
```diff
@@ -175,6 +176,7 @@ export const Link: FunctionComponent<{
   onClick?: (e?: KeyboardEvent) => void
 }>
 export const Modal: FunctionComponent<{
```
milan-bc, line 178: minor it could be pure functional component no need to use react `FunctionComponent` - but it's not breaker or stopper ;)

packages/blockchain-wallet-v4-frontend/src/components/Flyout/index.tsx
```diff
@@ -160,42 +153,41 @@ export const StickyHeaderFlyoutWrapper = styled(FlyoutWrapper)`
...
-                {/* Each child must be wrapped in FlyoutChild for transitioning to work */}
-                {children}
-              </ReactCSSTransitionGroup>
-            </FlyoutChildren>
-          </FlyoutModal>
-        )}
-      </Transition>
-    )
-  }
+const Flyout = ({ children, isOpen, ...props }: Props) => {
```
milan-bc, line 156: 👏

milan-bc (commented):
I left few comments in general great improvement I love it, let's double check error and slowness and then we can 🚀 it :)

packages/blockchain-wallet-v4-frontend/src/components/Alerts/template.tsx
pedroapfilho: How did this happen?

schnogz (approved):

PR #3085, opened 2021-04-13
feat(open banking): show error screen when a user cancels a deposit from their bank app, as well as when the retry times out

blockdylanb:
## Description (optional)
When creating a fiat deposit using Yapily/Open Banking we need to show an error screen when the user cancels the deposit from their bank app, as well as show a "timed out" screen when our long polling reaches it's limit without hearing back from the api.

## Testing Steps (optional)
- go to fiat wallet with GBP/EUR account
- create OB deposit
- click cancel on the modelo website


packages/blockchain-wallet-v4-frontend/src/data/components/brokerage/sagas.ts
```diff
@@ -313,7 +316,7 @@ export default ({
     try {
       const data = yield call(api.createFiatDeposit, amount, id, currency)
       if (partner === 'YAPILY') {
-        const order = yield retry(100, 10000, AuthUrlCheck, data.paymentId)
+        const order = yield retry(30, 10000, AuthUrlCheck, data.paymentId)
```
milan-bc, line 319: is this going to call API too often?

packages/blockchain-wallet-v4-frontend/src/data/components/brokerage/sagas.ts
```diff
@@ -325,7 +328,17 @@ export default ({
               dwStep: BankDWStepType.DEPOSIT_CONNECT
             })
           )
-          yield retry(100, 10000, ClearedStatusCheck, data.paymentId)
+          try {
+            const updatedOrder: SBTransactionType = yield retry(
+              30,
+              10000,
```
milan-bc, line 334: (minor) I would move all of those magic numbers in self explained constants we can reuse them on several places inside this file and maybe around the app :)

packages/blockchain-wallet-v4-frontend/src/scenes/Transactions/TransactionList/index.tsx
```diff
@@ -47,11 +47,12 @@ class TransactionList extends PureComponent<Props> {
                 currency={currency}
               />
             ) : 'priceFunnel' in tx ? (
-              <SwapOrderTx order={tx} coin={coin as CoinType} />
+              <SwapOrderTx key={i} order={tx} coin={coin as CoinType} />
```
milan-bc, line 50: maybe to use something else than `i`, like `tx.hash`?

milan-bc (approved):
LGTM, I left few minor comments

packages/blockchain-wallet-v4-frontend/src/data/components/brokerage/sagas.ts
blockdylanb: It's 1 call every 10 seconds for 30 total retries. That's 5 minutes. From my testing Yapily can take a 3-5 minutes to process an order and there's reason to believe it could take longer in a production environment.

packages/blockchain-wallet-v4-frontend/src/scenes/Transactions/TransactionList/index.tsx
blockdylanb: yeah that's what the first component in that list was using, but I think hash doesn't exist in the other components. I'll double check though.

Godspowerogun (approved):

packages/blockchain-wallet-v4-frontend/src/scenes/Transactions/TransactionList/index.tsx
blockdylanb: each tx has a unique `ts.id` so I used that instead.

PR #3088, opened 2021-04-13
feat(open banking): call stuff only in case that user exists

milan-bc:
## Description (optional)
This PR solve issue with calling endpoint in case that nabu user still do not exists

## Testing Steps (optional)
Create new wallet try to open buy/sell flow

### DIFF
--- a/packages/blockchain-wallet-v4-frontend/src/data/components/simpleBuy/sagas.ts
+++ b/packages/blockchain-wallet-v4-frontend/src/data/components/simpleBuy/sagas.ts
@@ -1292,10 +1292,7 @@ export default ({
   const showModal = function * ({ payload }: ReturnType<typeof A.showModal>) {
     const { cryptoCurrency, orderType, origin } = payload
     const latestPendingOrder = S.getSBLatestPendingOrder(yield select())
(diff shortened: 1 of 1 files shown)


packages/blockchain-wallet-v4-frontend/src/data/components/simpleBuy/sagas.ts
```diff
@@ -1292,17 +1292,18 @@ export default ({
...
       actions.modals.showModal('SIMPLE_BUY_MODAL', { origin, cryptoCurrency })
     )
     const fiatCurrencyR = selectors.core.settings.getCurrency(yield select())
     const fiatCurrency = fiatCurrencyR.getOrElse('USD')
 
     if (latestPendingOrder) {
+      const bankAccount = yield call(
+        getBankInformation,
+        latestPendingOrder as SBOrderType
+      )
```
milan-bc, line 1306: we should not call this one before since there might be a fresh wallet

blockdylanb: LGTM

blockdylanb (approved):

PR #3092, opened 2021-04-14
feat(coins): better formatting of coins removing if (coin) statement

milan-bc:
## Description (optional)
This PR is improvements and removing of nasty `if (coin)` stuff

## Testing Steps (optional)
make sure that all actions works properly, like sell/buy, swap

### DIFF
--- a/packages/blockchain-wallet-v4-frontend/src/data/coins/utils.ts
+++ b/packages/blockchain-wallet-v4-frontend/src/data/coins/utils.ts
@@ -41,15 +41,10 @@ export const generateInterestAccount = (
 
 export const generateProvisionalPaymentAmount = (
   coin: CoinType,
-  amount: number,
-  orgCoin?: CoinType
+  amount: number
 ): string | number => {
   if (coin === 'BTC' || coin === 'BCH') {
     return parseInt(convertStandardToBase(coin, amount))
-  }
-
-  if (coin === 'ETH' && orgCoin && orgCoin === 'USDT') {
-    return parseInt(convertStandardToBase(orgCoin, amount))
   }
 
   return convertStandardToBase(coin, amount)
--- a/packages/blockchain-wallet-v4-frontend/src/data/components/borrow/sagas.utils.ts
+++ b/packages/blockchain-wallet-v4-frontend/src/data/components/borrow/sagas.utils.ts
@@ -38,11 +38,7 @@ export default ({
     destination: string
   ): Generator<PaymentType | CallEffect, boolean, any> {
     let paymentError
-    const paymentAmount = generateProvisionalPaymentAmount(
-      payment.coin,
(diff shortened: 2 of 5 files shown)


packages/blockchain-wallet-v4-frontend/src/data/coins/utils.ts
```diff
@@ -41,16 +41,11 @@ export const generateInterestAccount = (
...
+  amount: number
 ): string | number => {
   if (coin === 'BTC' || coin === 'BCH') {
     return parseInt(convertStandardToBase(coin, amount))
   }
 
-  if (coin === 'ETH' && orgCoin && orgCoin === 'USDT') {
-    return parseInt(convertStandardToBase(orgCoin, amount))
-  }
-
```
schnogz, line 54: where is this logic now?

milan-bc: we do not need this logic since we now pass proper coin here, so before we did pass `payment.coin` which was in case of erc20 coin converted to ETH and we pass over ETH as coin parameter for all, like USDT and then our convertStandardToBase was in use for USDT on way that we pass `ETH` and `amount`, it was not ok because in that case formatting of USDT will be based on ETH, caused issue in first place, so this PR instead of having 3rd parameter in this function where we pass original coin we pass original coin as first one and we do not use `payment.coin` in such case, this will brings us to point that `convertStandardToBase` will now format price based on original coin.

pedroapfilho (approved):

packages/blockchain-wallet-v4-frontend/src/data/coins/utils.ts
schnogz: but there is not `parseInt` anymore for `ETH`. i think this is now going to return a `String`. Is this accounted for and tested?

schnogz (changes requested):
see response above

packages/blockchain-wallet-v4-frontend/src/data/coins/utils.ts
milan-bc: @schnogz I did tested this and it works as expected. Just a  more overview  as a hot fix I introduced this line

```
 if (coin === 'ETH' && orgCoin && orgCoin === 'USDT') {
    return parseInt(convertStandardToBase(orgCoin, amount))
  }
```
this was specially look into USDT case, but if you look further  next block is 

```
return convertStandardToBase(coin, amount)
```
which has been executed for all other coins including a ETH etc, I realise that `parseInt` was redundant there, so it works fine without it.

schnogz: thanks

schnogz (approved):

PR #3093, opened 2021-04-14
feat(exchange): adding login button on wallet login

pedroapfilho:
https://user-images.githubusercontent.com/13142568/114760347-95623180-9d35-11eb-89f4-6c00705714c6.mov


packages/blockchain-wallet-v4-frontend/src/layouts/Public/components/Footer/ExchangeLogin/index.tsx
```diff
@@ -0,0 +1,40 @@
+import React from 'react'
+import { FormattedMessage } from 'react-intl'
+import styled from 'styled-components'
+
+import { Icon, Link, Text } from 'blockchain-info-components'
+
+const ExchangePill = styled.div`
+  cursor: pointer;
```
milan-bc, line 8: I think cursor is not necessary since it's inside link already

milan-bc (approved):
LGTM I left one minor comment ;)

packages/blockchain-wallet-v4-frontend/src/layouts/Public/components/Footer/ExchangeLogin/index.tsx
pedroapfilho: I thought the same thing, but if I remove it, it doesn't work...

PR #3101, opened 2021-04-16
feat(buy): implemented logic to get in consideration proper limits based on product

milan-bc:
## Description (optional)
This PR add trading limits endpoint in use for proper finding what is min/max

## Testing Steps (optional)
Login to wallet try to buy DOT and enter 1, then click on buy min and try to buy, you should be able to buy it


packages/blockchain-wallet-v4/src/network/api/simpleBuy/types.ts
```diff
@@ -199,7 +199,11 @@ export type ISBBuyOrderType = {
   state: SBOrderStateType
   updatedAt: string
 }
-export type SBOrderActionType = 'BUY' | 'SELL'
+export enum OrderType {
+  BUY = 'BUY',
+  SELL = 'SELL'
+}
+export type SBOrderActionType = keyof typeof OrderType
```
milan-bc, line 206: created type out of enum and reused enum on all other logical places.

packages/blockchain-wallet-v4/src/network/api/simpleBuy/index.ts
```diff
@@ -420,6 +425,19 @@ export default ({
...
+    product: NabuCustodialProductType,
+    networkFee: CoinType,
+    side: SBOrderActionType
+  ): SwapUserLimitsType =>
+    authorizedGet({
+      url: nabuUrl,
+      endPoint: `/trades/limits?currency=${currency}&product=${product}&networkFee=${networkFee}&side=${side}`,
+      contentType: 'application/json',
+      ignoreQueryParams: true
+    })
```
milan-bc, line 439: My thought is to no matter as we call same and point business logic should remain in product so all related simpleBuy should be here etc...

packages/blockchain-wallet-v4/src/network/api/settingsComponent/types.ts
```diff
@@ -244,7 +245,7 @@ export type SBTransactionType = {
         status: 'UNCONFIRMED' | 'CONFIRMED'
         txHash: string
       }
-      type: 'DEPOSIT' | 'REFUNDED' | 'SELL'
+      type: 'DEPOSIT' | 'REFUNDED' | OrderType.SELL
```
milan-bc, line 248: we can make this more typed in some future PRs

packages/blockchain-wallet-v4-frontend/src/data/components/simpleBuy/sagas.ts
```diff
@@ -1377,14 +1379,24 @@ export default ({
...
+        limits = yield call(
+          api.getSBLimits,
+          currency,
+          ProductTypes.SIMPLEBUY,
+          cryptoCurrency,
+          side
+        )
+      } else {
+        limits = yield call(api.getSwapLimits, currency)
+      }
```
milan-bc, line 1399: we now have calling of same endpoint but with passing along diff parameters, once when we move SM calls from `packages/blockchain-wallet-v4-frontend/src/scenes/Home/Banners/index.tsx` then we can clean this up

packages/blockchain-wallet-v4-frontend/src/modals/SimpleBuy/EnterAmount/Checkout/index.tsx
```diff
@@ -39,8 +40,13 @@ class Checkout extends PureComponent<Props> {
       this.props.simpleBuyActions.fetchSDDEligible()
       this.props.simpleBuyActions.fetchSBCards()
       this.props.brokerageActions.fetchBankTransferAccounts()
-      this.props.simpleBuyActions.fetchLimits(this.props.fiatCurrency || 'USD')
     }
+    // we need to re-fetch limits for SB
```
schnogz, line 44: please explain "why" in the comment, we already know the "what" :)

packages/blockchain-wallet-v4-frontend/src/data/components/simpleBuy/sagas.ts
```diff
@@ -229,7 +231,7 @@ export default ({
             A.setStep({
               step: 'ENTER_AMOUNT',
               cryptoCurrency,
-              orderType: order.side || 'BUY',
+              orderType: order.side || OrderType.BUY,
```
blockdylanb, line 234: 👏🏻

packages/blockchain-wallet-v4-frontend/src/modals/SimpleBuy/EnterAmount/Checkout/index.tsx
milan-bc: oh yeah, because we call it here https://github.com/blockchain/blockchain-wallet-v4-frontend/pull/3101/files#diff-7eb86189b3d4261d7094ff70289228c91e6b3c3a976934f2e86254280aeff3f6L27 without additional parameters, will add it in comment ;)

packages/blockchain-wallet-v4/src/network/api/simpleBuy/index.ts
blockdylanb: my first thought was that these two requests could just be merged, but I see your logic. If we merged them into one call should the api endpoint live in the simple buy apis dir or the swap apis dir, or do we create a new "generic" dir for this one endpoint? In the end I think you're right that we just duplicate this in order to keep the business logic on the client and not over abstract the code. 👍🏻

blockdylanb: One improvement you could make to this is to make the arguments a single object called `data` and can be passed directly to the `authorizedGet` object and it'll turn them in to proper query params so your url can be way shorter.

```
const getSBLimits = (data: SwapUserLimitsType) =>
    authorizedGet({
      url: nabuUrl,
      endPoint: `/trades/limits`,
      contentType: 'application/json',
      data,
      ignoreQueryParams: true,
    })
```

blockdylanb (approved):

packages/blockchain-wallet-v4/src/network/api/simpleBuy/index.ts
milan-bc: I like this one, yeah can apply it

packages/blockchain-wallet-v4-frontend/src/modals/SimpleBuy/EnterAmount/Checkout/index.tsx
schnogz: thank you

milan-bc: @schnogz added ;)

PR #3109, opened 2021-04-19
Fix/prov payments

TheLeoB:
## Description (optional)
Fixes the issue where inputs for non-custodial sell/swap/interest weren't updating in real time, as the amount was being changed on the form. This resulted in fee estimates for BTC send sometimes being a lot lower than fees used during the transaction.

## Testing Steps (optional)

### DIFF
--- a/packages/blockchain-wallet-v4-frontend/src/data/components/simpleBuy/sagas.ts
+++ b/packages/blockchain-wallet-v4-frontend/src/data/components/simpleBuy/sagas.ts
@@ -939,8 +939,12 @@ export default ({
         Number(cryptoAmt),
         account.coin
       )
-      payment = yield payment.amount(paymentAmount)
-      payment = yield payment.build()
+      if (payment.coin === 'BTC' || payment.coin === 'BCH') {
+        payment = yield payment.build()
+      } else {
+        payment = yield payment.amount(paymentAmount)
+      }
+
       yield put(A.updatePaymentSuccess(payment.value()))
     } catch (e) {
       // eslint-disable-next-line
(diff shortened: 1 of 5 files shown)


packages/blockchain-wallet-v4-frontend/src/data/components/simpleBuy/sagas.ts
```diff
@@ -939,7 +939,12 @@ export default ({
         Number(cryptoAmt),
         account.coin
       )
-      payment = yield payment.amount(paymentAmount)
+      if (payment.coin === 'BTC' || payment.coin === 'BCH') {
```
pedroapfilho, line 942: Is this a good idea?

schnogz: @plondon did you have a look at this?  any other ideas?

TheLeoB: I changed this bit of code here to include all coins, (removing the BTC and BCH check), which fixed an issue for ETH but created another issue for XLM. When trying to send, I would see an error about reserve currency being greater than balance. 

https://github.com/blockchain/blockchain-wallet-v4-frontend/blob/86fbecd18239fdc28b7598c11f14ab5317da19aa/packages/blockchain-wallet-v4-frontend/src/data/components/swap/sagas.ts#L132

plondon (changes requested):
Before diving deeper, adding an initial stamp of 👎

totongcute (commented):
hi

PR #3113, opened 2021-04-21
feat(ob): fixed sizing of logo image

milan-bc:
## Description (optional)
This PR fix bank logo size on OB flow

## Testing Steps (optional)
Go over OB flow, try to Link a new bank, choose one and then you will see something like this

image

### DIFF
--- a/packages/blockchain-wallet-v4-frontend/src/modals/components/index.tsx
+++ b/packages/blockchain-wallet-v4-frontend/src/modals/components/index.tsx
@@ -71,6 +71,11 @@ const LinkOptionsWrapper = styled.div`
   justify-content: space-between;
   align-items: center;
 `
+const LogoImage = styled.img`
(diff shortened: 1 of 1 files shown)


schnogz (changes requested):
i feel like this entire file should be in `src/components/Brokerage`

milan-bc:
@schnogz that file is some common place we use it on SB and brokerage flows, maybe we can name it better than this.

blockdylanb:
> i feel like this entire file should be in src/components/Brokerage
> @schnogz that file is some common place we use it on SB and brokerage flows, maybe we can name it better than this.

This code was initially in brokerage, but the code is used in simple buy as well so I created the generic components dir with quite a bit of reusable components in it. The file is a little big but it made reusing this and other components so much nicer during development that I really think we should keep it this way.

blockdylanb (approved):

PR #3118, opened 2021-04-22
feat(ob): prevent infinite loading of QR code since order attributes are missing after repen of SB flow

milan-bc:
## Description (optional)
This PR improve loading of order details

## Testing Steps (optional)
create OB order close it after bank is linked, you will see banner that there is pending order, click on it and then QR screen will be stuck to infinite load since there are no order attribute details

### DIFF
--- a/packages/blockchain-wallet-v4-frontend/src/modals/SimpleBuy/OpenBankingConnect/index.tsx
+++ b/packages/blockchain-wallet-v4-frontend/src/modals/SimpleBuy/OpenBankingConnect/index.tsx
@@ -20,6 +20,12 @@ const Connect = (props: Props) => {
       )
     }
   }
+  useEffect(() => {
+    const { id } = props.order
+    if (id) {
+      props.simpleBuyActions.pollSBOrder(id)
+    }
+  }, [])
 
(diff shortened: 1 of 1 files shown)


milan-bc:
@blockdylanb check this one and let me know your thoughts

packages/blockchain-wallet-v4-frontend/src/modals/SimpleBuy/OpenBankingConnect/index.tsx
```diff
@@ -20,6 +20,12 @@ const Connect = (props: Props) => {
       )
     }
   }
+  useEffect(() => {
+    const { id } = props.order
+    if (id) {
+      props.simpleBuyActions.pollSBOrder(id)
+    }
+  }, [])
```
milan-bc, line 28: actually we might do not need this if we solve this one https://github.com/blockchain/blockchain-wallet-v4-frontend/blob/development/packages/blockchain-wallet-v4-frontend/src/data/components/simpleBuy/sagas.ts#L1323

blockdylanb (approved):

PR #3120, opened 2021-04-22
feat(simplebuy): fixed validation of min/max values and SDD flow

milan-bc:
## Description (optional)
This PR fix problem with min/max validation

## Testing Steps (optional)
Try to buy some crypto, enter $1 and then click on min


packages/blockchain-wallet-v4-frontend/src/modals/SimpleBuy/EnterAmount/Checkout/validation.tsx
```diff
@@ -121,13 +121,21 @@ export const getMaxMin = (
...
+          }
+
           if (!allValues) return defaultMax
           if (!method) return defaultMax
 
           let max = BigNumber.minimum(
             method.limits.max,
             isSddFlow ? Number(sddLimit.max) : pair.buyMax,
-            limitMaxAmount
+            isSddFlow ? Number(sddLimit.max) : limitMaxAmount
```
milan-bc, line 138: for SDD flow we still want to keep max of SDD max not limit max

schnogz (approved):

config/jest/jest.config.js
```diff
@@ -1,6 +1,7 @@
-import 'babel-polyfill'
-import Adapter from 'enzyme-adapter-react-16'
 import Enzyme from 'enzyme'
+import Adapter from 'enzyme-adapter-react-16'
+
+import "@babel/polyfill"
```
blockdylanb, line 4: usually the pollyfill be the first import in any file. Does this still work as intended?

milan-bc: this is what I get when save file our importing formatter, format it on this way.

PR #3139, opened 2021-04-27
Fix/export history

TheLeoB:
## Description (optional)
1. Export history from haskoin (bch/btc) returns unformatted values for fiat values. Added a function to format values that come back from haskoin.
2. Fixes issue where we were passing the derivations array to export bch history, instead of just the legacy address. We have to fish out the legacy address from the array and then pass it to the api call.
## Testing Steps (optional)


pedroapfilho (approved):

packages/blockchain-wallet-v4-frontend/src/modals/Transactions/DownloadTransactions/selectors.js
```diff
@@ -91,30 +91,34 @@ const getEthData = createSelector(
...
     selectors.core.wallet.getWallet,
+    selectors.core.settings.getCurrency,
     selectors.core.data.btc.getTransactionHistory
   ],
-  (wallet, dataR) => {
+  (wallet, currencyR, dataR) => {
+    const currency = currencyR.getOrElse('USD')
     const transform = data => {
-      const transformedData = map(tx => formatTxData(tx, 'BTC'), data)
+      const transformedData = map(tx => formatHaskoinData(tx, 'BTC', currency), data)
```
blockdylanb, line 100: this is totally fine, but if you might not get all the typescripty goodness using the ramda `map`. Might be better off just using good ol `data.map(...)`

blockdylanb (approved):

pedroapfilho (approved):

packages/blockchain-wallet-v4-frontend/src/modals/Transactions/DownloadTransactions/selectors.js
plondon: JS file still so more work needed to get to TS anyway

packages/blockchain-wallet-v4/src/redux/data/bch/sagas.ts
```diff
@@ -147,7 +147,13 @@ export default ({ api }: { api: APIType }) => {
       yield put(A.fetchTransactionHistoryLoading())
       const currency = yield select(selectors.settings.getCurrency)
       if (address) {
-        const convertedAddress = convertFromCashAddrIfCashAddr(address)
+         // TODO: SEGWIT remove w/ DEPRECATED_V3
+         // remove address.length check, all
+         // wallets will have a derivations array
+        const bchLegacyAddress = prop('address', address.length === 2 && address.find( add => add.type === 'legacy'))
```
plondon, line 153: formatting is weird here

PR #3171, opened 2021-05-10
feat(buy): we do not need to convert to base since amount is already converted to base

milan-bc:
## Description (optional)
This PR fix max/min amount in c

## Testing Steps (optional)
Try to buy some crypto with and without method selected, max amount should be aligned to API response not minor one

image

image

### DIFF
--- a/packages/blockchain-wallet-v4-frontend/src/modals/SimpleBuy/EnterAmount/Checkout/validation.tsx
+++ b/packages/blockchain-wallet-v4-frontend/src/modals/SimpleBuy/EnterAmount/Checkout/validation.tsx
@@ -122,12 +122,8 @@ export const getMaxMin = (
             )
           }
 
-          const defaultLimitMaxAmount = convertBaseToStandard(
-            'FIAT',
-            limitMaxAmount
-          )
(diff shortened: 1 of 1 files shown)


packages/blockchain-wallet-v4-frontend/src/modals/SimpleBuy/EnterAmount/Checkout/validation.tsx
```diff
@@ -122,12 +122,8 @@ export const getMaxMin = (
...
           }
 
-          const defaultLimitMaxAmount = convertBaseToStandard(
-            'FIAT',
-            limitMaxAmount
-          )
-          if (Number(defaultMax.FIAT) > Number(defaultLimitMaxAmount)) {
-            defaultMax.FIAT = defaultLimitMaxAmount
+          if (Number(defaultMax.FIAT) > limitMaxAmount) {
+            defaultMax.FIAT = String(limitMaxAmount)
```
milan-bc, line 126: amount is already converted to base since it came like that from BE we do not need to convert it again

blockdylanb (approved):
LGTM

PR #3183, opened 2021-05-13
feat(buy): fixed override of min/max amount if it's not changed

milan-bc:
## Description (optional)
Fix min/max on fresh new account

## Testing Steps (optional)
Create new wallet and the try to buy (SDD or non SDD) should apply on both


pedroapfilho (approved):

blockdylanb (approved):
Tested locally, works like a charm.

Aivolt1 (commented):
.

PR #3188, opened 2021-05-13
fix(deposits): default to USD when adding a new bank from deposit

blockdylanb:
## Description (optional)
Fiat deposit -> add bank screen wasn't showing up because of the fiatcurrency prop. Set a getOrElse default to USD to solve the issue for USD, EUR, and GBP.


dsisto-bc (approved):
LGTM!

schnogz (approved):

PR #3189, opened 2021-05-14
fix(xlm): increase tx fetch interval, stop stream reconnects after failure

schnogz:
## Description
in order to both clean up errors in the console and reduce the number of calls we make to full nodes, do the following:
1) if we fail to subscribe to stream for an account, no longer retry to resub
2) tx fetching interval is somehow tied to the reconnect interval constant, bump that up from 30 seconds to 6 minutes


blockdylanb (approved):
👍🏻

PR #3195, opened 2021-05-17
fix(metadata): check the correct and incorrect encryption key

plondon:
## Description (optional)
Metadata fetch fails because of an incorrectly derived private key used for decrypting the payload.

## Testing Steps (optional)

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -8190,7 +8190,7 @@ eslint-config-airbnb-base@^14.2.1:
     object.assign "^4.1.2"
     object.entries "^1.1.2"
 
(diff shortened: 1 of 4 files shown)


yarn.lock
```diff
@@ -8190,7 +8190,7 @@ eslint-config-airbnb-base@^14.2.1:
     object.assign "^4.1.2"
     object.entries "^1.1.2"
 
-eslint-config-airbnb@^18.2.1:
+eslint-config-airbnb@18.2.1:
```
plondon, line 8193: Sorry, ignore these

mpfluger (approved):

PR #3199, opened 2021-05-18
fix(metadata): naming and refactor of parse, use padded as default

plondon:
## Description (optional)

## Testing Steps (optional)

### DIFF
--- a/packages/blockchain-wallet-v4/src/types/KVStoreEntry.js
+++ b/packages/blockchain-wallet-v4/src/types/KVStoreEntry.js
@@ -174,6 +174,7 @@ export const extractResponse = curry((encKey, encKeyUnpadded, res) => {
           compose(JSON.parse, decrypt(encKey), prop('payload'))(res)
         )
       } catch (e) {
+        console.error(e)
         console.info(
           'LOG: trying unpadded enc key',
           compose(JSON.parse, decrypt(encKeyUnpadded), prop('payload'))(res)
--- a/packages/blockchain-wallet-v4/src/types/KVStoreEntry.js
+++ b/packages/blockchain-wallet-v4/src/types/KVStoreEntry.js
@@ -33,7 +33,7 @@ export const magicHash = KVStoreEntry.define('magicHash')
 export const address = KVStoreEntry.define('address')
 export const signKey = KVStoreEntry.define('signKey')
 export const encKeyBuffer = KVStoreEntry.define('encKeyBuffer')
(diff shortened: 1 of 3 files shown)


packages/blockchain-wallet-v4/src/types/KVStoreEntry.js
```diff
@@ -163,36 +161,41 @@ export const verifyResponse = curry((address, network, res) => {
...
     // TODO: remove when redux/core is moved to frontend
     if (window.logLevel === 'verbose' && res.type_id !== -1) {
       // eslint-disable-next-line no-console
-      console.info(
-        'LOG: ',
-        encKey
-          ? compose(decrypt(encKey), prop('payload'))(res)
-          : compose(BufferToString, B64ToBuffer, prop('payload'))(res)
-      )
+      try {
```
mpfluger, line 171: Is this just for debugging or will it be part of the final PR

plondon: It logs when users login with that special log-level

plondon: I could probably remove it now since the reason we put it there was to debug what was probably this issue.

schnogz (approved):

packages/blockchain-wallet-v4/src/types/KVStoreEntry.js
mpfluger: Since we're failing hard again on failure and this should fix the issue I'd say we can kill it again. But I'm fine either way.

packages/blockchain-wallet-v4/src/types/KVStoreEntry.js
```diff
@@ -33,7 +33,7 @@ export const magicHash = KVStoreEntry.define('magicHash')
 export const address = KVStoreEntry.define('address')
 export const signKey = KVStoreEntry.define('signKey')
 export const encKeyBuffer = KVStoreEntry.define('encKeyBuffer')
-export const encKeyBufferPadded = KVStoreEntry.define('encKeyBufferPadded')
+export const encKeyBufferPadded = KVStoreEntry.define('encKeyBufferUnpadded')
```
paulo-bc, line 36: Shouldn't this be 
`export const encKeyBufferUnpadded = KVStoreEntry.define('encKeyBufferUnpadded') `

PR #3204, opened 2021-05-20
feat(translations): update react-intl, add formatjs/cli

schnogz:
## Description (optional)
- Updates `react-intl` from v2 to v5
- Use `@formatjs/cli` to extract messages to `en.json` file without the need to build the app.  Meaning `yarn release` is now ~3 minutes faster
- Removes no longer needed libs `babel-plugin-react-intl` and `react-intl-translations-manager` 
- Fixes previously mismatched id/messages and removes use of `FormattedHTMLMessage` as in v3 of `react-intl` its just `FormattedMessage`


packages/blockchain-wallet-v4-frontend/package.json
```diff
@@ -117,7 +115,7 @@
     "react-csv": "2.0.3",
     "react-dom": "16.14.0",
     "react-dropzone": "7.0.1",
-    "react-intl": "2.8.0",
+    "react-intl": "5.17.7",
```
milan-bc, line 118: 🚀

PR #3209, opened 2021-05-21
feat(recurring buys): implemented home page banner

milan-bc:
## Description (optional)
This PR implements banner for home page

## Testing Steps (optional)
Log in on wallet and then you will see banner


schnogz:
please wrap rendering logic in a feature flag.  I added the new section and a flag for this feature in wallet-options. eg. 

```
featureFlags: {
  recurringBuys: true
}
```

schnogz (changes requested):
see comment above

milan-bc:
@schnogz I added feature flag,

PR #3212, opened 2021-05-21
fix(analytics): adding id to segment identify

pedroapfilho:
## Description (optional)

## Testing Steps (optional)

### DIFF
--- a/packages/blockchain-wallet-v4-frontend/src/middleware/analyticsMiddleware.ts
+++ b/packages/blockchain-wallet-v4-frontend/src/middleware/analyticsMiddleware.ts
@@ -204,16 +204,12 @@ const analytics = queuevent<AnalyticsKey, AnalyticsPayload>({
 
     const analyticsURL = `${options.domains.api}/events/publish`
 
+    const nabuId = rawEvents.find((event) => event.payload.id)?.payload.id || null
+
     const context: { traits: { nabu_id: string | null } } = {
       traits: {
-        nabu_id: null
+        nabu_id: nabuId
       }
-    }
-
-    const nabuId = rawEvents.find((event) => event.payload.id)?.payload.id
-
-    if (nabuId) {
-      context.traits.nabu_id = nabuId
     }
 
     const events = rawEvents.map((event) => {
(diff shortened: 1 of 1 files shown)


packages/blockchain-wallet-v4-frontend/src/middleware/analyticsMiddleware.ts
```diff
@@ -195,14 +195,26 @@ const simpleBuyPaymentTypeDictionary = (rawPaymentType: SBPaymentTypes) => {
...
     const analyticsURL = `${options.domains.api}/events/publish`
 
-    const id = v4()
+    const context: { traits: { nabu_id: string | null } } = {
+      traits: {
+        nabu_id: null
+      }
+    }
+
+    const nabuId = rawEvents.find((event) => event.payload.id)?.payload.id
```
blockdylanb, line 213: if payload is undefined this will throw, can you guarantee it's always there? Also it seems like your `find()` is just looking for the first event with a payload id. Is that what you want it to do?

packages/blockchain-wallet-v4-frontend/src/middleware/analyticsMiddleware.ts
```diff
@@ -195,14 +195,26 @@ const simpleBuyPaymentTypeDictionary = (rawPaymentType: SBPaymentTypes) => {
...
+
+    if (nabuId) {
+      context.traits.nabu_id = nabuId
```
blockdylanb, line 216: you could default nabuId to null on line 213 and then simply remove the `if` logic here.

packages/blockchain-wallet-v4-frontend/src/middleware/analyticsMiddleware.ts
```diff
@@ -195,14 +195,22 @@ const simpleBuyPaymentTypeDictionary = (rawPaymentType: SBPaymentTypes) => {
...
   queueCallback: async (rawEvents) => {
     const res = await fetch('/wallet-options-v4.json')
     const options = await res.json()
 
     const analyticsURL = `${options.domains.api}/events/publish`
 
-    const id = v4()
+    const nabuId = rawEvents.find((event) => event.payload.id)?.payload.id || null
+
+    const context: { traits: { nabu_id: string | null } } = {
```
milan-bc, line 209: maybe to make type from this one like

```
type ContextType = { traits: { nabu_id: string | null } } 
```

and then use it, but it's really minor

milan-bc (approved):
lgtm

PR #3213, opened 2021-05-21
fix(sb): fixing the minimum amount for simple buy

pedroapfilho:


packages/blockchain-wallet-v4-frontend/src/modals/SimpleBuy/EnterAmount/Checkout/validation.tsx
```diff
@@ -166,7 +166,7 @@ export const getMaxMin = (
           let max = BigNumber.minimum(
             method.limits.max,
             isSddFlow ? sddLimit.max : pair.buyMax,
-            isSddFlow ? sddLimit.max : convertBaseToStandard('FIAT', limitMaxAmount, false)
+            isSddFlow ? sddLimit.max : limitMaxAmount
```
blockdylanb, line 169: @milan-bc you made this change 4 days ago, but it seems to be double converting the limitMaxAmount which causes the min amount to be $500 in the UI. Is there something that we're missing here we should know about?

milan-bc: hm, I changed it since it was minor string and we need to convert it back, we have to check is amount double precise and then just not convert it

milan-bc (approved):

PR #3218, opened 2021-05-24
Replace .error color with readable color

dcondrey:
Error messages in darkMode are nearly unreadable.   Lightened color to increase contrast against dark background.

## Description (optional)
Replaced .error color in DarkMode to increase contrast against dark background.  Existing color definition was nearly unreadable.

Preview


schnogz (approved):
Thanks for this!

PR #3221, opened 2021-05-25
fix(change): variable change addr types

plondon:
## Description (optional)

## Testing Steps (optional)


schnogz (commented):
Why the `eslint-disable`'s?  If its just for a clean PR review i understand but we should make it a point to re-enable and fix after the review

mpfluger (approved):

PR #3238, opened 2021-06-02
fix(interest): remove payment.build for custodial payments

TheLeoB:
## Description (optional)
Fixes FWLT-1417. 
I believe `payment.build()` won't work if 'from' account is custodial, because payment will try to build using unspents, but they will be missing from a custodial send. `Payment.build()` was added recently to make sure we were calculating network fees in real time for BTC send, accounting for any additional inputs being added to the transaction as the amount was changing dynamically.
I added a check to only update the payment with `payment.build()` if it's not a custodial wallet. 

## Testing Steps (optional)


milan-bc (approved):
This issue fixed EDD interest bug, thanks @TheLeoB

PR #3265, opened 2021-06-08
fix(cache): track incorrect caches

plondon:
## Description (optional)
Silently track if the account xpub receive/change cache is incorrect.


packages/blockchain-wallet-v4-frontend/src/middleware/analyticsMiddleware/index.ts
```diff
@@ -19,16 +19,17 @@ import { convertBaseToStandard } from 'data/components/exchange/services'
 import { ModalNamesType } from 'data/types'
 
 const analyticsMiddleware = () => (store) => (next) => (action) => {
+  const state = store.getState()
+  const nabuId = state.profile.userData.getOrElse({})?.id
+  const id = state.walletPath.wallet.guid
+
```
pedroapfilho, line 25: This should be inside your case

pedroapfilho (changes requested):
If that is not inside your case, then you'll have the `store.getState` running on every action, and it can make the app slow

pedroapfilho (approved):

mpfluger (approved):

PR #3274, opened 2021-06-11
hotfix(link_bank): adding walletCurrency as default for checking the currency

pedroapfilho:


schnogz (approved):

packages/blockchain-wallet-v4-frontend/src/modals/Brokerage/Banks/AddBankStatus/selectors.ts
```diff
@@ -6,17 +6,10 @@ import { RootState } from 'data/rootReducer'
...
-  return lift(
-    (
-      bankStatus: ExtractSuccess<typeof bankStatusR>,
-      bankCredentials: ExtractSuccess<typeof bankCredentialsR>
-    ) => ({
-      bankStatus,
-      bankCredentials
-    })
-  )(bankStatusR, bankCredentialsR)
+  return lift((bankStatus: ExtractSuccess<typeof bankStatusR>) => ({
```
blockdylanb, line 10: Bankcredentials was being included here just to display the bank name on the bank link success screen, but was causing that screen to just spin on loading forever for Yodlee bank accounts. So I just removed it.

PR #3276, opened 2021-06-11
fix(interest): form error if account balance is zero

TheLeoB:
## Description (optional)
Interest deposit form throws an error if user 
1) has a private key account with 0 balance selected as their deposit account 
2) and then enters an amount into the field

To fix this issue:
1) If source account balance <= 0, don't create a payment
2) Disable amount field if chosen deposit account has 0 balance

## Testing Steps (optional)


packages/blockchain-wallet-v4-frontend/src/modals/Interest/DepositForm/template.success.tsx
```diff
@@ -121,7 +121,9 @@ const DepositForm: React.FC<InjectedFormProps<{ form: string }, Props> & Props>
   const lockUpDuration = interestLimits[coin]?.lockUpDuration || 7200
   const lockupPeriod = lockUpDuration / 86400
   const maxDepositFiat = maxFiat(depositLimits.maxFiat, walletCurrency)
-
+  const accountBalance =
+    values && values?.interestDepositAccount && values.interestDepositAccount.balance
```
blockdylanb, line 125: you may want to put and `|| something` at the end here.

PR #3277, opened 2021-06-11
fix(brokerage): fixes infinite spinner in general settings add bank button click and infinite spinner when successfully linking a bank

blockdylanb:
## Description (optional)
There are two similar but unrelated issues. 
1. When clicking "Add a bank" from the general settings, the modal has an infinite loading screen.
2. When you link a bank through settings/simple buy/deposit the loading spinner spins infinitely and never transitions to the "Successfully linked bank" screen

## Testing Steps (optional)
Try to link a bank from general settings, and also link a bank from simple buy and it should get all the way to the success screen.

### DIFF
--- a/packages/blockchain-wallet-v4-frontend/src/modals/Brokerage/Banks/AddBankStatus/selectors.ts
+++ b/packages/blockchain-wallet-v4-frontend/src/modals/Brokerage/Banks/AddBankStatus/selectors.ts
@@ -3,20 +3,17 @@ import { lift } from 'ramda'
 import { ExtractSuccess } from 'blockchain-wallet-v4/src/types'
 import { selectors } from 'data'
 import { RootState } from 'data/rootReducer'
+import { OBType } from 'data/types'
 
 export const getData = (state: RootState) => {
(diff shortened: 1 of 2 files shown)


pedroapfilho:
Shouldn't you change the `fiatCurrency` default to `USD` on `packages/blockchain-wallet-v4-frontend/src/data/components/withdraw/reducers.ts` too?

pedroapfilho (approved):
Other than the default currency for withdraw, it looks good!

milan-bc (approved):
LGTM, also do agree with @pedroapfilho that we need default value for fiatCurrency

blockdylanb:
> Shouldn't you change the `fiatCurrency` default to `USD` on `packages/blockchain-wallet-v4-frontend/src/data/components/withdraw/reducers.ts` too?

Actually I don't think we should, I don't like we have default values like that in the first place, but changing it might affect other places in the app. Also, this code change makes it so that the modal prefers the fiat from the userData limits which should always be there if the user even sees the "add a bank" button on the settings page.

PR #3280, opened 2021-06-12
Modernize createPaymentTransactionNew API

DesWurstes:
## Description

The Ledger createPaymentTransactionNew API was updated. Currently, using the legacy interface prints a warning. The new interface is also makes code easier to read.

https://github.com/LedgerHQ/ledgerjs/blob/7184a62c3469949743efa3c4fc2b93293add5e04/packages/hw-app-btc/src/Btc.js#L137-L157

## Testing Steps
Any tests that run the signer.


plondon:
Thanks for the PR!

@schnogz or @TheLeoB  do you have a ledger you can test with?

schnogz:
What version of the app requires this @DesWurstes ?  We are currently on `@ledgerhq/hw-app-btc": "4.33.7`

DesWurstes:
Oh no. Yes this is for a much later version. It would not be worth it for you to update v4 to v5 only to make the code more readabke.

PR #3290, opened 2021-06-16
feat(analytics): adding events for withdrawal_v1

pedroapfilho:


packages/blockchain-wallet-v4-frontend/src/modals/Settings/TradingLimits/template.success.tsx
```diff
@@ -376,9 +376,12 @@ const Template: React.FC<Props> = (props) => {
...
                 fullwidth
                 nature='primary'
-                onClick={() =>
+                onClick={() => {
                   analyticsActions.logEvent(INTEREST_EVENTS.SETTINGS.SUPPLY_INFORMATION)
-                }
+                  /* interestActions.handleWithdrawalSupplyInformation({
+                    origin: 'Settings'
+                  }) */
+                }}
```
milan-bc, line 384: maybe delete this one since it's not in use?

pedroapfilho: This one is actually just not used because this is going to be on `v2`, that's why it's commented

milan-bc (approved):

PR #3294, opened 2021-06-18
Redux toolkit

blockdylanb:
## Description (optional)
Proof of concept PR for what the process will look like to migrate all of the codebase to redux-toolkit

## Testing Steps (optional)


packages/blockchain-wallet-v4-frontend/src/data/components/brokerage/slice.ts
```diff
@@ -0,0 +1,115 @@
...
+    setDWStep: (state, action: PayloadAction<BrokerageDWStepPayload>) => {
+      state.dwStep = action.payload.dwStep
+      if (action.payload.dwStep === BankDWStepType.DEPOSIT_METHODS) {
+        state.addNew = action.payload.addNew || false
+      }
+    },
+    showModal: (state, action: PayloadAction<{origin: BrokerageModalOriginType, modalType: ModalNamesType}>) => {},
+    handleDepositFiatClick: (state, action: PayloadAction<WalletFiatType>) => {
+      state.fiatCurrency = action.payload
+    },
```
milan-bc, line 89: I like this, what I might do is to extract all multi line functions to be stand alone functions above this definition and then here just pass them, might be easier for maintenance I think

blockdylanb: ah so something like 
```
// maybe in another file?
const handleDepositFiatClick = (state, action: PayloadAction<WalletFiatType>) => {
    state.fiatCurrency = action.payload
}

// in slice.ts
import { handleDepositFiatClick } from './another file'
const brokerageSlice = createSlice({
    name: 'brokerage',  
    initialState: INITIAL_STATE, 
    reducers: {
        handleDepositFiatClick,
        ....
    }
})
```

packages/blockchain-wallet-v4-frontend/src/data/components/brokerage/slice.ts
```diff
@@ -0,0 +1,115 @@
...
+const brokerageSlice = createSlice({
+  name: 'brokerage',
+  initialState: INITIAL_STATE,
+  reducers: {
+    createFiatDeposit: () => {},
+    deleteSavedBank: (state, action: PayloadAction<BankTransferAccountType['id']>) => {},
+    handleWithdrawClick: (state, action: PayloadAction<WalletFiatType>) => {},
+    fetchBankLinkCredentials: (state, action: PayloadAction<WalletFiatType>) => {},
+    fetchBankLinkCredentialsLoading: (state) => {
+      state.bankCredentials = Remote.Loading
```
blockdylanb, line 42: These are basically the same reducers we've always had, but notice how we can modify the state directly in here and not run into issues with redux. This is because redux-toolkit uses the immer library behind the scenes to handle creating a new state object for us. 

Now instead of defining a new action and action type in another file, RTK does it for us. At the bottom of this file you can see there is `brokerageSlice.actions` which is all the generated actions such as in this example `brokerageSlice.actions.fetchBankLinkCredentialsLoading` and when we want the action type, you just call `brokerageSlice.actions.fetchBankLinkCredentialsLoading.type` such as in `sagaRegister.ts`

packages/blockchain-wallet-v4-frontend/src/data/components/brokerage/types.ts
```diff
@@ -188,81 +186,3 @@ export type BrokerageState = {
   fiatCurrency: WalletFiatType | undefined
   redirectBackToStep: boolean
 }
-
-interface FetchBankLinkCredentialsType {
```
blockdylanb, line 192: This was very redundant boilerplate that we needed, but since reducers/actions/actionTypes are created by RTK, we can just delete these as the library handles all the proper types!

packages/blockchain-wallet-v4-frontend/src/data/components/simpleBuy/sagas.ts
```diff
@@ -983,10 +982,10 @@ export default ({ api, coreSagas, networks }: { api: APIType; coreSagas: any; ne
         )
       case SBPaymentTypes.LINK_BANK:
         yield put(
-          actions.components.brokerage.showModal(
-            BrokerageModalOriginType.ADD_BANK,
-            fiatCurrency === 'USD' ? 'ADD_BANK_YODLEE_MODAL' : 'ADD_BANK_YAPILY_MODAL'
-          )
+          actions.components.brokerage.showModal({
```
blockdylanb, line 985: One snag I ran into is that RTK reducers enforce a `(state, payload) => {}` style for all reducers/actions so we can't pass in multiple arguments like we have done for the `showModal()` action in the past. Now the arguments must be passed as a single object (the payload), but this is was really easy to change throughout the app.

packages/blockchain-wallet-v4-frontend/src/store/index.js
```diff
@@ -90,9 +91,20 @@ const configureStore = async function () {
   })
   const persistWhitelist = ['session', 'preferences', 'cache']
 
-  // TODO: remove getStoredStateMigrateV4 someday (at least a year from now)
-  const store = createStore(
-    connectRouter(history)(
+  const store = configureStore({
```
blockdylanb, line 94: I just did a basic amount of effort here to get RTK working, but there's more we could do here. Converting to TS file, and using RTKs defaultMiddleware are a couple wins.

packages/blockchain-wallet-v4-frontend/src/data/components/brokerage/slice.ts
vaske: not necessary to be in separate file, but yeah something like that ;)

milan-bc: not necessary to be in separate file, but yeah something like that ;)

dsisto-bc:
👍

packages/blockchain-wallet-v4-frontend/src/store/index.js
```diff
@@ -46,12 +48,9 @@ const devToolsConfig = {
   serialize: serializer
 }
 
-const configureStore = async function () {
+const configuredStore = async function () {
   const history = createHashHistory()
   const sagaMiddleware = createSagaMiddleware()
-  const composeEnhancers = window.__REDUX_DEVTOOLS_EXTENSION_COMPOSE__
-    ? window.__REDUX_DEVTOOLS_EXTENSION_COMPOSE__(devToolsConfig)
-    : compose
```
plondon, line 54: Does devtools still work?

packages/blockchain-wallet-v4-frontend/src/store/index.js
```diff
@@ -89,10 +88,33 @@ const configureStore = async function () {
     reauthenticate
   })
   const persistWhitelist = ['session', 'preferences', 'cache']
-
-  // TODO: remove getStoredStateMigrateV4 someday (at least a year from now)
-  const store = createStore(
-    connectRouter(history)(
+  // console.log('fooooo', [
```
plondon, line 91: do we need this comment?

packages/blockchain-wallet-v4-frontend/src/store/index.js
blockdylanb: yeah part of redux-toolkit is that devtools is add/configured automatically.

packages/blockchain-wallet-v4-frontend/src/store/index.js
blockdylanb: whoops... deleting.

pedroapfilho (approved):

PR #3295, opened 2021-06-18
fix(update-mnemonic-call): wrap call in try catch to handle error better

TheLeoB:
## Description (optional)
Separated update mnemonic code into its own saga
Wrapped saga in a try/catch so login continues even if update mnemonic fails

## Testing Steps (optional)


blockdylanb (approved):
LGTM

PR #3303, opened 2021-06-21
Feat/release sso

TheLeoB:
## Description (optional)
We reset a user's kyc when they recover their wallet with the phrase

This ensures that if the reset call fails, user is logged out of the wallet

## Testing Steps (optional)


packages/blockchain-wallet-v4-frontend/src/data/auth/sagas.ts
```diff
@@ -285,6 +285,8 @@ export default ({ api, coreSagas }) => {
       yield call(coreSagas.data.xlm.fetchLedgerDetails)
       yield call(coreSagas.data.xlm.fetchData)
 
+      yield call(authNabu, fromRestoredFlow)
```
milan-bc, line 288: Does this works for brand new wallet user do not exists?

packages/blockchain-wallet-v4-frontend/src/data/modules/profile/sagas.ts
```diff
@@ -261,13 +262,28 @@ export default ({ api, coreSagas, networks }) => {
...
+    try {
+      yield put(A.resetUserKycLoading())
+      yield call(coreSagas.kvStore.walletCredentials.fetchMetadataWalletCredentials)
+      const retailToken = yield call(generateRetailToken)
+      const userId = (yield select(selectors.core.kvStore.userCredentials.getUserId)).getOrElse(
+        null
+      )
+      const lifetimeToken = (yield select(
+        selectors.core.kvStore.userCredentials.getLifetimeToken
+      )).getOrFail()
```
milan-bc, line 274: (minor) since this part is only needed in case that `userId` exists, you an easily move it inside try block

milan-bc (commented):
Great work so far @TheLeoB I see there are some build errors with dependencies it's something we should investigate

PR #3306, opened 2021-06-22
feat(kyc): send US state format in US-XX rather than XX

milan-bc:
## Description (optional)
We send now US state in proper format which API side expects

## Testing Steps (optional)
create a new wallet go over KYC and enter address in US


pedroapfilho (approved):
LGTM

PR #3311, opened 2021-06-24
fix(analytics): adding origin dictionaries for KYC

pedroapfilho:


milan-bc (approved):

packages/blockchain-wallet-v4-frontend/src/components/Announcements/Static/index.js
```diff
@@ -13,17 +13,15 @@ const Wrapper = styled.div`
...
-  background: ${props => props.theme.blue600};
+  background: ${(props) => props.theme.blue600};
   overflow: hidden;
   ${media.tablet`
     display: none;
   `};
 `
 
 class StaticAnnouncementsContainer extends React.PureComponent {
-  state = {}
```
milan-bc, line 24: we use state later need to have constructor

PR #3317, opened 2021-06-28
fix(kyc-reset): if user has just been created dont reset kyc

TheLeoB:
## Description (optional)
POST /users endpoint is built on getOrCreate model and is also being called during KYC reset, returning lifetimeToken & userId, however some of the users have been stick to non-custodial accounts only. We need to indicate if the account was created or not for frontends to check if we need to skip the /kyc/reset endpoint.

response from /users now includes a `created` param indicating whether or not user has just been created. 
## Testing Steps (optional)


milan-bc (approved):

Huper7777 (commented):
Thanks

PR #3325, opened 2021-06-30
Feat/release sso

TheLeoB:
## Description (optional)
1) Removes unused 'from restored param'
2) cleans up UI when scanning code from mobile app for login
3) Sets restoreFromMetadataSuccess if user recovers without nabuID being reset

## Testing Steps (optional)

### DIFF
--- a/packages/blockchain-wallet-v4-frontend/src/data/auth/sagas.ts
+++ b/packages/blockchain-wallet-v4-frontend/src/data/auth/sagas.ts
@@ -530,7 +530,7 @@ export default ({ api, coreSagas }) => {
         } catch (e) {
           // if it fails with user already being reset, shuold be allowed
           // to continue with flow
-          if (e.description === 'User reset in progress') {
+          if (e.status === 409) {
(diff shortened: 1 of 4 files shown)


packages/blockchain-wallet-v4-frontend/src/data/auth/sagas.ts
```diff
@@ -530,7 +525,7 @@ export default ({ api, coreSagas }) => {
         } catch (e) {
           // if it fails with user already being reset, shuold be allowed
           // to continue with flow
-          if (e.description === 'User reset in progress') {
+          if (e.status === 409) {
```
milan-bc, line 528: would be cool to move 409 to be in const we can define more HTTP status codes and reuse when ever we need ;)

milan-bc (approved):
I left one minor comment, rest looks neat

packages/blockchain-wallet-v4-frontend/src/data/auth/sagas.ts
TheLeoB: ohh, good idea, thanks.

PR #3367, opened 2021-07-19
feat(swap): fixed issue with circle size when text is too wide

milan-bc:
## Description (optional)
This tiny PR solve CSS issue

## Testing Steps (optional)
Login with wallet which contains swap account with huge amount of money make sure that circle looks as expected
image


packages/blockchain-wallet-v4-frontend/src/components/Form/CoinAccountListOption/index.tsx
```diff
@@ -62,6 +62,7 @@ const CircleBorder = styled.div`
   justify-content: center;
   align-items: center;
   width: 24px;
+  min-width: 24px;
```
milan-bc, line 65: this is actually the fix

pedroapfilho (approved):

PR #3375, opened 2021-07-20
fix(swap): amount form

plondon:
## Description (optional)
If you are on the summary page of swap and return to the `swapAmount` form, we will rerun initAmountForm with a  provisionalPayment with value = `0`. This fixes that problem. This problem is live in prod as well.

## Testing Steps (optional)


plondon:
https://blockchain.atlassian.net/browse/FWN-613

schnogz (approved):

PR #3414, opened 2021-07-29
fix(sell): show all cryptos

plondon:
## Description (optional)
It's fixed in the other PR I made but so we can unblock users who want to sell

## Testing Steps (optional)


milan-bc (approved):
lgtm

PR #3420, opened 2021-07-29
fix(simple buy): convert all min amounts from standard to base before displaying in the UI

blockdylanb:
## Description (optional)
Issue with min buy showing up as $685.00 instead of $6.85 because the value isn't being converted from base amount to standard amount.

image

### DIFF
--- a/packages/blockchain-wallet-v4-frontend/src/modals/SimpleBuy/EnterAmount/Checkout/validation.tsx
+++ b/packages/blockchain-wallet-v4-frontend/src/modals/SimpleBuy/EnterAmount/Checkout/validation.tsx
@@ -242,7 +242,7 @@ export const getMaxMin = (
             isSddFlow ? method.limits.min : limitMinAmount
           ).toString()
 
-          const minFiat = !limitMinChanged ? convertBaseToStandard('FIAT', min) : min
+          const minFiat = convertBaseToStandard('FIAT', min)
           const minCrypto = getQuote(quote.pair, quote.rate, 'FIAT', minFiat)
 
           return { CRYPTO: minCrypto, FIAT: minFiat }


packages/blockchain-wallet-v4-frontend/src/modals/SimpleBuy/EnterAmount/Checkout/validation.tsx
```diff
@@ -242,7 +242,7 @@ export const getMaxMin = (
             isSddFlow ? method.limits.min : limitMinAmount
           ).toString()
 
-          const minFiat = !limitMinChanged ? convertBaseToStandard('FIAT', min) : min
+          const minFiat = convertBaseToStandard('FIAT', min)
```
milan-bc, line 245: I think this fix was previously introduced to prevent some edge case, we just need to test all possible edge cases.

pedroapfilho: Do you remember which one @milan-bc ?

blockdylanb: The only time `limitMinChanged` is set to true is when `limits?.minOrder` exists and is greater than `pair.buyMin`. From what I could see limits.minOrder is coming back from the api in the standard format (life would be easier if it was just base format). Where this might break is if the api returns base format sometimes and not other times or if `isSddFlow` is true and that causes something weird to happen.

schnogz (approved):

packages/blockchain-wallet-v4-frontend/src/modals/SimpleBuy/EnterAmount/Checkout/validation.tsx
milan-bc: @pedroapfilho nah, they sometimes pops up like that after someone reported it

PR #3439, opened 2021-08-03
Hotfix(Swap): max button doesn't fill in proper amount

blockdylanb:
## Description (optional)
In swap if you click the "swap max" button while the form is in fiat mode the conversion from fiat -> crypto often results in incorrect amounts due to rounding which results in the amount being more than the user actually has, when this amount is submitted the api errors and users get stuck in a strange state. On mobile clients, when the user clicks the "swap max" the form automatically switches to crypto mode and inputs the users actual maximum amount. I fix this issue on web by doing the same as mobile clients.

Screen Shot 2021-08-03 at 3 50 37 PM

## Testing Steps (optional)


schnogz (approved):

pedroapfilho (approved):

schnogz:
will just release this on monday as planned.

PR #3465, opened 2021-08-12
fix(chat): load chat at app level to ensure active chats always persist

schnogz:
## Description
- users with active chats had the chat disappear when changing routes in the app. moving higher in the stack ensures it is always loaded
- refactors chat widget a bit to be more performant
- converts data/session to redux-toolkit


packages/blockchain-wallet-v4-frontend/src/data/session/slice.ts
```diff
@@ -0,0 +1,25 @@
...
+import { SessionStateType } from './types'
+
+const initialState: SessionStateType = {}
+
+const sessionSlice = createSlice({
+  initialState,
+  name: 'session',
+  reducers: {
+    removeSession: (state, action: PayloadAction<string>) => {
+      state = dissoc(action.payload, state.sessions)
```
blockdylanb, line 13: This is totally fine, but you also don't have return a new object here if you don't want. You can modify the state object like `delete state[action.payload]` but understandable if you don't want to do it that way since `delete` has it's own problems.

packages/blockchain-wallet-v4-frontend/src/data/session/slice.ts
```diff
@@ -0,0 +1,25 @@
...
+    },
+    saveSession: (state, action: PayloadAction<{ [key: string]: string }>) => {
+      state = mergeRight(state, action.payload)
```
blockdylanb, line 16: Same thing here, but looking into it, you don't actually know what the key/value is going to be so this seems right.

blockdylanb (approved):

PR #3478, opened 2021-08-17
fix(charts): svg width param cannot be negative

blockdylanb:
## Description (optional)
The charts were doing some math to set the width of the chart, but some of the first renders were rendering a negative number which is not allowed for svg element. I just used `Math.abs()` to make sure it's always a positive number.

image


schnogz (approved):
great catch 👏

PR #3582, opened 2021-09-09
Refactor (Price Chart) :  Price Chart Redux-Toolkit migrate

mitchmalinin:


packages/blockchain-wallet-v4-frontend/src/data/components/priceChart/slice.ts
```diff
@@ -0,0 +1,40 @@
...
+    },
+    initialized: {
+      prepare: (coin: CoinType, time: TimeRange) => ({ payload: { coin, time } }),
+      reducer: (state, action: PayloadAction<any>) => {
+        state.coin = action.payload.coin
+        state.time = action.payload.time
+      }
+    },
+    timeClicked: {
+      prepare: (time: TimeRange) => ({ payload: { time } }),
```
plondon, line 28: Thanks for this!

What does prepare do here?

mitchmalinin: Yea no problem! Not sure if you know this already but the prepare callback is a built in function inside slice that allows the payload to be formatted. Since there is not really action creators with this style of creating actions/reducers - the prepare is able to act as an action creator. I need to use the prepare for the "initialize" action/reducer because the payload is 2 arguments. 

As for the other ones it seems like there is a weird relationship between this reducer and the preferences reducer. Inside the preferences reducer the "timeClicked" and "coinClicked" actions are called and they expect a payload to have a corresponding "time" and "coin" key in the payload object. I tried to just reformat the preferences reducer, but I was running into typescript issues. So instead of changing all that, I just formatted the "timeClicked"/"coinClicked" payloads to have those time and coin values! If you find a better way please let me know!

schnogz (approved):

PR #3595, opened 2021-09-10
feat(deposit): payment method limits double check

milan-bc:
## Description (optional)
Fixed issue when trying to deposit fiat

## Testing Steps (optional)

### DIFF
--- a/packages/blockchain-wallet-v4-frontend/src/modals/Brokerage/Banks/Deposit/EnterAmount/template.success.tsx
+++ b/packages/blockchain-wallet-v4-frontend/src/modals/Brokerage/Banks/Deposit/EnterAmount/template.success.tsx
@@ -122,7 +122,7 @@ const LimitSection = ({ fiatCurrency, paymentMethods }) => {
     (method) => method.type === SBPaymentTypes.BANK_TRANSFER
   )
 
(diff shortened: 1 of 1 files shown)


packages/blockchain-wallet-v4-frontend/src/modals/Brokerage/Banks/Deposit/EnterAmount/template.success.tsx
```diff
@@ -122,7 +122,7 @@ const LimitSection = ({ fiatCurrency, paymentMethods }) => {
     (method) => method.type === SBPaymentTypes.BANK_TRANSFER
   )
 
-  if (bankTransfer?.limits) {
+  if (bankTransfer?.limits?.daily) {
```
milan-bc, line 125: @blockdylanb I double checked with Lorenzo this part was not changed but I do not see anything daily related just min/max inside limits

blockdylanb: I think we shouldn't have the `?.daily` here on second thought. It's in prod api now, but they want to remove it so then this code will never execute.

blockdylanb:
Closing this as it's issues are covered in another PR.

PR #3598, opened 2021-09-10
Refactor(Settings): Redux tool-kit migrate 

mitchmalinin:


packages/blockchain-wallet-v4-frontend/src/data/components/settings/slice.ts
```diff
@@ -0,0 +1,39 @@
...
+const initialState: SettingsState = {
+  productsEligibility: Remote.NotAsked
+}
+
+const settingsSlice = createSlice({
+  initialState,
+  name: 'settings',
+  reducers: {
+    fetchProductsEligibility: () => {},
+    fetchProductsEligibilityFailure: (state: SettingsState, action) => {
```
blockdylanb, line 16: Wow, great work on this, thanks! This is looking almost perfect. One change I would make here is that redux-toolkit has a `PayloadAction` type that you can use. So this line would look like this:
```
fetchProductsEligibilityFailure: (state: SettingsState, action: PayloadAction<ProductEligibility[]>) => {
```
Then you can actually delete the `FetchProductsEligibilityFailure` interface in /components/settings/types.ts since RTK handles all of the payload type validation for you!

blockdylanb (changes requested):

schnogz (approved):

PR #3605, opened 2021-09-14
feat(signup): fetch settings after we update wallet currency

milan-bc:
## Description (optional)
Set default currency based on country

## Testing Steps (optional)
Create a new wallet from some EU country default currency should be EUR

### DIFF
--- a/packages/blockchain-wallet-v4-frontend/src/data/auth/sagas.ts
+++ b/packages/blockchain-wallet-v4-frontend/src/data/auth/sagas.ts
@@ -277,8 +277,7 @@ export default ({ api, coreSagas, networks }) => {
 
         yield put(actions.modules.settings.updateCurrency(currency, true))
         yield put(actions.core.settings.setCurrency(currency))
-        // fetch settings again
-        // yield call(coreSagas.settings.fetchSettings)
+
         if (!isAccountReset) {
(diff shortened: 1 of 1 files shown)


packages/blockchain-wallet-v4-frontend/src/data/auth/sagas.ts
```diff
@@ -275,9 +275,10 @@ export default ({ api, coreSagas, networks }) => {
         const countryCode = country || 'US'
         const currency = guessCurrencyBasedOnCountry(countryCode)
 
+        yield put(actions.modules.settings.updateCurrency(currency, true))
         yield put(actions.core.settings.setCurrency(currency))
         // fetch settings again
-        yield call(coreSagas.settings.fetchSettings)
+        // yield call(coreSagas.settings.fetchSettings)
```
schnogz, line 281: remove the two commented out lines

schnogz (changes requested):

packages/blockchain-wallet-v4-frontend/src/data/auth/sagas.ts
milan-bc: @schnogz yeah done ✅

schnogz (approved):

PR #3606, opened 2021-09-14
feat(interest): extended BigNumbers Exponential values to keep sending amounts in correct format

milan-bc:
## Description (optional)
This PR fix withdrawal formatting of huge amounts

## Testing Steps (optional)
Try to transfer from interest flow some big amount in crypto like 1000 or something


plondon (changes requested):
18 is ETHs precision....

plondon:
This is also linked to development instead of the release/v4.55 branch.

PR #3629, opened 2021-09-15
feat(country on signup): fixed issue with select country to show state in case of US

milan-bc:
## Description (optional)
Fixed issue with proper select country

## Testing Steps (optional)
Go over signup and change country to be US state should be visible


schnogz (approved):

milan-bc:
we don't need this one since we will get those changes back from ongoing release

PR #3645, opened 2021-09-20
feat(signup): set currency based on country after signup with country

milan-bc:
## Description (optional)
Based on country we do set user's currency

## Testing Steps (optional)
Try over signup process change country and later currency should be based on country

### DIFF
--- a/packages/blockchain-wallet-v4-frontend/src/data/auth/sagas.ts
+++ b/packages/blockchain-wallet-v4-frontend/src/data/auth/sagas.ts
@@ -276,7 +276,6 @@ export default ({ api, coreSagas, networks }) => {
         const currency = guessCurrencyBasedOnCountry(countryCode)
 
         yield put(actions.modules.settings.updateCurrency(currency, true))
-        yield put(actions.core.settings.setCurrency(currency))
 
         if (!isAccountReset) {
           yield put(actions.router.push('/verify-email-step'))
--- a/packages/blockchain-wallet-v4-frontend/src/data/auth/sagas.ts
+++ b/packages/blockchain-wallet-v4-frontend/src/data/auth/sagas.ts
@@ -275,9 +275,9 @@ export default ({ api, coreSagas, networks }) => {
         const countryCode = country || 'US'
         const currency = guessCurrencyBasedOnCountry(countryCode)
 
+        yield put(actions.modules.settings.updateCurrency(currency, true))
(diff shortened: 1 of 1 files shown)


packages/blockchain-wallet-v4-frontend/src/data/auth/sagas.ts
```diff
@@ -275,9 +275,9 @@ export default ({ api, coreSagas, networks }) => {
         const countryCode = country || 'US'
         const currency = guessCurrencyBasedOnCountry(countryCode)
 
+        yield put(actions.modules.settings.updateCurrency(currency, true))
         yield put(actions.core.settings.setCurrency(currency))
-        // fetch settings again
-        yield call(coreSagas.settings.fetchSettings)
+
```
plondon, line 280: Doesn't updateCurrency call setCurrency?

milan-bc: It does, but without this like I run into NaN issue
image

plondon: Why do you need to call `yield put(actions.core.settings.setCurrency(currency))`

packages/blockchain-wallet-v4-frontend/src/data/auth/sagas.ts
```diff
@@ -326,6 +326,7 @@ export default ({ api, coreSagas, networks }) => {
         // store initial address in case of US state we add prefix
         const userState = country === 'US' ? `US-${state}` : state
         yield call(api.setUserInitialAddress, country, userState)
+        yield call(coreSagas.settings.fetchSettings)
```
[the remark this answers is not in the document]
milan-bc: originally it was added to be fetched for first time login on where we do set currency, right after that, I moved it here, because if it stay on that line it will fetch non updated state and currency will remain as default USD

packages/blockchain-wallet-v4-frontend/src/data/auth/sagas.ts
milan-bc: Let me try removing it, if all works as expected we can remove it.

milan-bc: great call, removed.

plondon (approved):

PR #3685, opened 2021-09-30
feat(analytics): adding analytics id to cookies

pedroapfilho:


packages/blockchain-wallet-v4-frontend/src/scenes/app.tsx
```diff
@@ -143,6 +143,9 @@ const App = ({
 }
 
 const mapStateToProps = (state) => ({
+  apiUrl: selectors.core.walletOptions.getDomains(state).getOrElse({
+    api: 'https://api.dev.blockchain.info'
```
schnogz, line 147: probably want to prefer prod on fallback.

schnogz (changes requested):

packages/blockchain-wallet-v4-frontend/src/scenes/app.tsx
pedroapfilho: hmm, good point!

PR #3690, opened 2021-09-30
fix(circular deps): resolve all circular dependencies

schnogz:
## Description 
- resolves all circular dependencies currently in app
- adds pre-push hook to check for new introductions of circular dependencies
- disables `import/no-extraneous-dependencies` eslint rule


plondon (approved):
noice

PR #3732, opened 2021-10-13
fix(buysell): use preferredFiatTradingCurrency from userData when checking buy sell eligibility instead of user view currency

blockdylanb:
## Description (optional)
When a user changes their wallet view currency to something like the Brazilian Real, and then opens the buy sell flow. The app requests payment methods with that currency which we don't support and so the app errors. We have a relatively new set of userData which has the users eligible currency that we get from the api. We should use this currency instead as it's is much more reliable. 

## Testing Steps (optional)


diyartofeeq:
DIYAR TOFEEQ

diyartofeeq:
@diyartofeeq

diyartofeeq:
diyarmohammed161@gmail.com

PR #3825, opened 2021-11-05
feat(seamless limits): swap moved to RTK and initial swap error messages

milan-bc:
## Description (optional)
This is PR to convert swap flow to use RTK

## Testing Steps (optional)
Test swapping with T3 and T2 wallets


packages/blockchain-wallet-v4-frontend/src/modals/Swap/EnterAmount/Checkout/validation.tsx
```diff
@@ -17,6 +17,7 @@ export const getMaxMin = (
   BASE: SwapAccountType,
   COUNTER: SwapAccountType
 ) => {
+  // eslint-disable-next-line default-case
```
schnogz, line 20: can we just add `default: break` and then not disable the rule?

packages/blockchain-wallet-v4-frontend/src/modals/model.ts
```diff
@@ -0,0 +1,15 @@
...
+    case current?.daily?.effective:
+      return current.daily
+    case current?.monthly?.effective:
+      return current.monthly
+    case current?.yearly?.effective:
+      return current.yearly
+    default:
+      return undefined
+  }
+}
```
schnogz, line 15: this is not the correct place for this file IMO.  a model file here should be useable by most all modals.   Perhaps we create a new service folder named `custodial` and put it in there?  thoughts @blockdylanb ?

schnogz: having it as a service then means sagas could use it as well and not be importing a model from modals which makes little sense.

schnogz (changes requested):
Thanks for Swap converting to redux-toolkit!  We need to be extra diligent and making sure this refactor was done correctly.  Please test as many use/edge cases regarding swap as you can.  Also make sure QA is notified in next release ticket to focus on Swap.

packages/blockchain-wallet-v4-frontend/src/modals/Swap/EnterAmount/Checkout/validation.tsx
milan-bc: sure, will change

packages/blockchain-wallet-v4-frontend/src/modals/model.ts
milan-bc: I'm up to any of those, my idea/intention was to keep it on common place and re-use it in several others since we literately need it for every flow.

blockdylanb: I think a service file makes sense in this case. It would be nice to have fewer model files spread throughout the code. I'd personally prefer to have a small amount of model files close to the top of the file hierarchy. I think it's ok that they are bigger files because as a developer you mostly access them by clicking the "Go to Definition" option in the code editor vs scrolling around in that file looking for things.

schnogz (approved):

Donone19 (approved):

PR #3831, opened 2021-11-08
fix(nft-sell-orders): makerRelayerFees fixed

jmichel-bc:
## Description (optional)

## Testing Steps (optional)

### DIFF
--- a/packages/blockchain-wallet-v4/src/redux/payment/nfts/index.ts
+++ b/packages/blockchain-wallet-v4/src/redux/payment/nfts/index.ts
@@ -7,7 +7,7 @@ import {
   _atomicMatch,
   _authorizeOrder,
   _buyOrderValidationAndApprovals,
-   _cancelOrder,
+  _cancelOrder,
   _makeBuyOrder,
   _makeMatchingOrder,
   _makeSellOrder,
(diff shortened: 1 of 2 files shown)


packages/blockchain-wallet-v4/src/redux/payment/nfts/index.ts
```diff
@@ -21,6 +21,9 @@ import {
 
 export const cancelNftListings = async (asset: NftAsset, signer: Signer) => {
   // TODO: on front end maybe worth having a way for users to select which one of their sell orders they want to cancel and then input the order directly to this function.
+  if (!asset.sell_orders) {
```
plondon, line 24: I think sell_orders is an array, so you may want to check !asset.sell_orders.length

packages/blockchain-wallet-v4/src/redux/payment/nfts/utils.ts
```diff
@@ -1072,7 +1076,7 @@ export async function _makeSellOrder({
     basePrice: new BigNumber(basePrice.toString()),
     calldata,
     englishAuctionReservePrice: reservePrice ? new BigNumber(reservePrice.toString()) : undefined,
-    exchange: '0x7Be8076f4EA4A4AD08075C2508e481d6C946D12b',
+    exchange: '0x7Be8076f4EA4A4AD08075C2508e481d6C946D12b'.toLowerCase(),
```
plondon, line 1079: Can you add a comment here since this isn't usual practice

plondon (changes requested):
small things

packages/blockchain-wallet-v4/src/redux/payment/nfts/utils.ts
jmichel-bc: I will update it to be hardcoded to lower case instead, my bad for the hacky fix!

plondon: All good!

packages/blockchain-wallet-v4/src/redux/payment/nfts/index.ts
Donone19: Thanks.

PR #3883, opened 2021-11-19
chore(yarn): upgrading to yarn 3

blockdylanb:
## Description (optional)
This PR fixes some issues we've had with using yarn workspaces. Now instead of manually editing package.json to add/remove packages you can use `yarn workspace blockchain-wallet-v4-frontend add date-fns` for example. 

When pulling this code down it's probably best to run `yarn clean` first to clear out node_modules and then run `yarn` to reinstall everything with the new set up. You may run into an error because your local yarn is on an older version, if so run `yarn set version berry` for yarn to upgrade itself to the newer version.

## Testing Steps (optional)


package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "blockchain-wallet-v4",
+  "name": "blockchain-wallet-v4-base",
```
blockdylanb, line 2: @schnogz the base package.json and the package.json in `/packages/blockhain-wallet-v4` had the same name and this caused a name conflict error in yarn. Can you see any issues coming up in the ci with changing the base package name?

package.json
```diff
@@ -56,7 +56,7 @@
     "ci:test:build": "yarn wsrun --serial test:build",
     "ci:test:core:components": "yarn wsrun --serial --exclude-missing ci:test",
     "ci:test:frontend": "yarn workspace blockchain-wallet-v4-frontend ci:test:frontend",
-    "clean": "cross-env yarn wsrun clean && rimraf build && rimraf coverage && rimraf dist && rimraf *.log && rimraf node_modules",
```
blockdylanb, line 59: when running `yarn clean` it was throwing an error (`No matches found: "*.log"`) so I removed this to make it work correctly.

package.json
schnogz: hmm, im not sure... will pull it down and test today

PR #3937, opened 2021-12-06
feat(nux complete your profile): feature flag and banner has been implemented

milan-bc:
## Description (optional)
This PR implements complete your profile banner, adding also feature flag and logic to calculate current progress step

## Testing Steps (optional)
after logging with non T2 user you should see something like this
image


package.json
```diff
@@ -190,6 +190,7 @@
     "node-polyfill-webpack-plugin": "1.1.4",
     "prettier": "2.3.2",
     "prettier-config-constellation": "0.0.5",
+    "react-circular-progressbar": "2.0.4",
```
blockdylanb, line 193: I think this should go in the `package/blockchain-wallet-v4-frontend/package.json` instead of this one and in the dependencies instead of devDependencies.

packages/blockchain-wallet-v4-frontend/src/components/CircularProgressBar/index.tsx
```diff
@@ -0,0 +1,22 @@
+import React from 'react'
+import { buildStyles, CircularProgressbarWithChildren } from 'react-circular-progressbar'
+
+import 'react-circular-progressbar/dist/styles.css'
+
+const ProgressCircle = ({ children, percentage, strokeWidth }) => {
+  return (
+    <CircularProgressbarWithChildren
+      styles={buildStyles({
+        backgroundColor: '#0C6CF2',
```
blockdylanb, line 10: possible to get these from the theme?

packages/blockchain-wallet-v4-frontend/src/scenes/Home/Banners/CompleteYourProfile/index.tsx
```diff
@@ -0,0 +1,134 @@
...
+    padding: 10px;
+  `}
+`
+
+const CompleteYourProfile = ({ buySellActions, data, fiatCurrency, verifyIdentity }: Props) => {
+  useEffect(() => {
+    buySellActions.fetchCards(false)
+    buySellActions.fetchPaymentMethods(fiatCurrency)
+    buySellActions.fetchBalance({ skipLoading: true })
+  }, [])
```
blockdylanb, line 64: should probably put `fiatCurrency` in the array here.

schnogz (approved):

packages/blockchain-wallet-v4-frontend/src/components/CircularProgressBar/index.tsx
milan-bc: well I might need direct import of color from theme, no other way to pass it over since this is not styled component :/

milan-bc: yeah you are right there is a hook useTheme in styled components which helps to use theme outside of styled components which is a case over here ;)

PR #3956, opened 2021-12-09
feat(fab-flyout): add fab trade flyout

jjBlockchain:
## Description (optional)
* Add FAB trade flyout


packages/blockchain-wallet-v4-frontend/src/layouts/Wallet/MenuTop/template.tsx
```diff
@@ -23,6 +23,12 @@ const Header = (props: OwnProps) => {
     props.sessionActions.logout()
   }, [])
 
+  const fabCallback = useCallback(() => {
+    props.modalActions.showModal('TRADE_MODAL', {
```
milan-bc, line 27: (minor) - you can import ModalName and use it here like `ModalName.TRADE_MODAL` rather then just a pure string

packages/blockchain-wallet-v4-frontend/src/modals/Trade/index.tsx
```diff
@@ -0,0 +1,192 @@
...
+      <FormattedMessage
+        id='modals.trade.receive.description'
+        defaultMessage='Cash Out to Your Bank Account'
+      />
+    ),
+    handleClick: () => {},
+    header: () => <FormattedMessage id='buttons.withdraw' defaultMessage='Withdraw' />,
+    iconName: 'withdraw'
+  }
+]
```
milan-bc, line 107: descriptions and headers are just formatted messages you do not need to make it as function could be just 

```
description:
      <FormattedMessage
        id='modals.trade.receive.description'
        defaultMessage='Cash Out to Your Bank Account'
      />,
...
header: <FormattedMessage id='buttons.withdraw' defaultMessage='Withdraw' />,
```

PR #3971, opened 2021-12-14
fix(transactions): filter out pending_confirmation and expired from t…

blockdylanb:
…he activity feed list

## Description (optional)

## Testing Steps (optional)


schnogz:
Let's discuss these changes.  Specifically I'm fine with removing `EXPIRED` trades but `PENDING_CONFIRMATION` is a bigger one.  Is `PENDING_CONFIRMATION` also applicable to on-chain transactions?  if so, I dont think we should make this change.

schnogz (approved):

PR #4043, opened 2021-12-21
fix(CON-2584): fix no quote found on selling trades

seanjameshan:
## Description (optional)
fixes no quote found error

## Testing Steps (optional)
both buy and sell


schnogz (approved):

Sinzu179 (commented):
Recovery

PR #4069, opened 2022-01-07
feat(signup) add captcha checks to wallet creation flows

schnogz:
## Description
To combat botting, we are adding hidden captcha checks to wallet creation flows.  This adds code to pass captcha back to wallet insert API method in both new wallet creation and wallet recovery via mnemonic phrase.


packages/blockchain-wallet-v4-frontend/src/data/auth/types.ts
```diff
@@ -95,20 +99,6 @@ export enum UserType {
...
-    ip_address: string
-  }
-  confirmation_required?: boolean
-  requester?: {
-    browser: string
-    country_code: string
-    ip_address: string
-  }
-  success: boolean
-}
```
schnogz, line 111: type was never used

PR #4110, opened 2022-01-21
feat(buyer hand hold): implemented new logic for dispatching events and added proper events for peak sheet

milan-bc:
## Description (optional)

## Testing Steps (optional)


packages/blockchain-wallet-v4-frontend/src/data/analytics/analytics.ts
```diff
@@ -0,0 +1,72 @@
...
+    console.error(e)
+  }
+}
+
+const analytics = queuevent<AnalyticsKey, AnalyticsValue>({
+  queueCallback,
+  queueName: QUEUE_NAME
+})
+
+export default analytics
```
milan-bc, line 72: at the moment we will have both files here and on old location until we migrate all events to new approach

packages/blockchain-wallet-v4-frontend/src/data/analytics/types.ts
```diff
@@ -0,0 +1,64 @@
...
+
+type PeekSheetViewedPayloadAction = {
+  key: AnalyticsKey.PEEKSHEET_VIEWED
+  properties: PeakSheetCurrentStep
+}
+
+type PeekSheetSelectionClickedPayloadAction = {
+  key: AnalyticsKey.PEEKSHEET_SELECTION_CLICKED
+  properties: PeakSheetCurrentStep & PeakSheetButtonClick
+}
```
milan-bc, line 57: thus way we have super typed stuff, so each key have dedicated properties type

packages/blockchain-wallet-v4-frontend/src/data/analytics/types/onboardingAndVerification.ts
```diff
@@ -0,0 +1,264 @@
+// Onboarding and Verification Events
+export enum Events {
+  ONBOARDING_ADDRESS_ENTERED = 'Address Information Entered',
```
blockdylanb, line 3: Not a show stopper, but I wonder if we really need these event keys as an enum? Since they're only ever used as the key to a specific type couldn't we save some lines of code by removing the enum and just putting the key string directly in the event type? 
ex.
```
type EmailVerificationRequestAction = {
  key: 'Email Verification Requested'
  properties: {
    origin: OriginProductFlow
  }
}
```
It should still be type safe and we should still have autocomplete I believe.

blockdylanb (approved):

packages/blockchain-wallet-v4-frontend/src/data/analytics/types/onboardingAndVerification.ts
milan-bc: Actually we use them when we cll event itself from any location, we can do it like you said but with current implementation we use enum and then you can use it like `AnalyticsKey.PEEKSHEET_SELECTION_CLICKED` anywhere in the app rather than to use just strings around.

PR #4115, opened 2022-01-25
fix(nfts): allow for big eth amounts

rsteubs-bc:
## Description (optional)

Converting to full string since BigNumber forces scientific notation, but opensea doesn't want
(ex. auto converts to 1e27 when we actually want '1000000000000000000000000000')
 
basePrice: BigNumber | string
extra: BigNumber | string

### DIFF
--- a/packages/blockchain-wallet-v4/src/redux/payment/nfts/utils.ts
+++ b/packages/blockchain-wallet-v4/src/redux/payment/nfts/utils.ts
@@ -1065,7 +1065,7 @@ export async function _makeSellOrder({
   quantity: number
   startAmount: number
   waitForHighestBid: boolean
-}): Promise<any> {
+}): Promise<UnhashedOrder> {
   // todo: re-implement this later:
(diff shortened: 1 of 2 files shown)


packages/blockchain-wallet-v4/src/redux/payment/nfts/utils.ts
```diff
@@ -1065,7 +1065,7 @@ export async function _makeSellOrder({
   quantity: number
   startAmount: number
   waitForHighestBid: boolean
-}): Promise<UnhashedOrder> {
+}): Promise<any> {
```
plondon, line 1068: why the change to any?

plondon:
Does this also fix the form issue? When entering 9999?

packages/blockchain-wallet-v4/src/redux/payment/nfts/utils.ts
rsteubs-bc: This is just for testing, because we're going from BigNumber to string, in unhashed order the baseprice type is a BigNumber

PR #4116, opened 2022-01-25
fix(buysell): fixes issue where min amount was incorrect when buying …

blockdylanb:
…denominated in crypto instead of fiat
image

## Description (optional)

## Testing Steps (optional)


milan-bc (approved):
lgtm, hate that we have to patch every now and then but that's how it is...

schnogz (approved):

PR #3966, opened 2021-12-13
fix(wallet): account derivations

plondon:
## Description (optional)
Some users have reported incomplete account derivations or missing derivations list. This PR finds those accounts on login and fixes them as long as user does not have a 2nd password.

## Testing Steps (optional)
Detail the steps required for the reviewer(s) to verify and test these changes.

TODO
- [x] test multiple accounts w/ bad derivations
- [x] check that the derivation is correct for account

PR #4133, opened 2022-01-28
fix(derivations): apply replenishDerivations fix whenever derivations array size is not 2.

paulo-bc:
##  Description (optional)

In addition to the fix implemented in #3966, we would like to also apply replenishDerivations fix whenever derivations array size is not 2.
This resolved an bug where there is only one derivation in the derivations array.

## Testing Steps (optional)

### DIFF
--- a/packages/blockchain-wallet-v4/src/redux/wallet/sagas.ts
+++ b/packages/blockchain-wallet-v4/src/redux/wallet/sagas.ts
@@ -326,7 +326,7 @@ export default ({ api, networks }) => {
     const wallet = yield select(S.getWallet)
     const accounts = Wallet.selectHDAccounts(wallet)
     const accountsWithMissingDerivations = accounts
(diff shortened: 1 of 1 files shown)


packages/blockchain-wallet-v4/src/redux/wallet/sagas.ts
```diff
@@ -326,7 +326,7 @@ export default ({ api, networks }) => {
     const wallet = yield select(S.getWallet)
     const accounts = Wallet.selectHDAccounts(wallet)
     const accountsWithMissingDerivations = accounts
-      .filter((acct) => acct.derivations.size === 0)
+      .filter((acct) => acct.derivations.size !== 2)
```
paulo-bc, line 329: @plondon should we do `acct.derivations.size < DERIVATION_LIST.length` like you first proposed?

PR #4144, opened 2022-02-01
chore(analytics): moved withdrawal events to new analytics tracking

milan-bc:
## Description (optional)
Withdrawal events added

## Testing Steps (optional)


schnogz (dismissed):

packages/blockchain-wallet-v4-frontend/src/middleware/analyticsMiddleware/index.ts
```diff
@@ -1788,107 +1788,6 @@ const analyticsMiddleware = () => (store) => (next) => (action) => {
...
-            withdrawal_method: withdrawalMethod
-          },
-          traits: {
-            email,
-            nabuId,
-            tier
-          }
-        })
-        break
-      }
```
blockdylanb, line 1891: 👏🏻

packages/blockchain-wallet-v4-frontend/src/modals/Brokerage/Banks/Withdraw/EnterAmount/index.tsx
```diff
@@ -148,6 +172,7 @@ const EnterAmountContainer = (props: Props) => {
           crossBorderLimits={crossBorderLimits}
           formErrors={formErrors}
           formActions={props.formActions}
+          onMaxButtonClicked={handleMaxButtonClicked}
```
blockdylanb, line 175: Given that there are so many params to the enter amount screen, and also that the analytics call is pretty static, I think it would be better (and a lot less code for you) if we just move the `handleMaxButtonClicked` method inside of the enter amount component.

blockdylanb (changes requested):

packages/blockchain-wallet-v4-frontend/src/modals/Brokerage/Banks/Withdraw/EnterAmount/index.tsx
milan-bc: I disagree because in that case I have to pass 2 more props to `EnterAmount` because it doesn't have clue about entered amount nor withdrawal method :/

blockdylanb: Yeah I see what you mean. Good call!

blockdylanb (approved):

PR #4155, opened 2022-02-03
feat(wallet): changing more labels from `wallet` to `account` wording

cmonti-bc:
## Description (optional)
More labels that were found with `wallet` wording that should change into `account` wording.


jtk-bc (approved):

packages/blockchain-wallet-v4-frontend/src/assets/locales/index.d.ts
```diff
@@ -2914,4 +2906,4 @@ type MessagesType = {
 }
 
 export type MessageIdType = keyof MessagesType
-export type DefaultMessageType = MessagesType[MessageIdType]
+export type DefaultMessageType = MessagesType[MessageIdType]
```
gsas-bc, line 2909: Missing new line?

gsas-bc (dismissed):

packages/blockchain-wallet-v4-frontend/src/assets/locales/index.d.ts
cmonti-bc: Great catch, fixed.

schnogz (changes requested):
I think these changes are too broad.  We still need to use the term Wallet in many of these contexts since they are using the Wallet product.   @amadeo-bc thoughts?

amadeo-bc:
Reopening. I need to look at this before we release unified signup.

PR #4159, opened 2022-02-04
feat(analytics): automated tracking with typwriter

blockdylanb:
## The Problem
We currently have several issues related to our segment tracking events setup.
- Large Switch/Case statements that listen to redux actions for firing off events.
- Dependence on creating new redux actions for every event we want to track
- Large amounts of typescript boilerplate code for event types
- No easy way to see/fix segment violations
- It's difficult to to be aware of new events that were added or existing events that have been changed.

A lot of these issues are solved with the [Typewriter](https://www.npmjs.com/package/typewriter) library from Segment, but initially we weren't able to use it because it depends on Segments `analytics.js` to send tracking events to segment but we proxy all of our events through our own API. But even if we can't use analytics.js, the schema and tracking event methods that Typewriter generates would still solve almost all of our problems. 

## The Solution
After installing typewriter and pulling down the schema and generated tracking event methods, the only remaining part was figuring out how to hijack `window.analytics` to direct the tracking event data to our api instead of to segments servers. I ended up going with a redux middleware because I needed the ability to dispatch redux actions so I could call our existing `trackEvent()` saga which adds some extra info to each event. The middleware is what made the whole thing come together. Now typewriter looks for `window.analytics` and gets my stubbed version back which, when called, fires our normal `trackEvent` redux action with all the proper (typewriter validated) event information!


packages/blockchain-wallet-v4-frontend/src/store/index.js
```diff
@@ -42,6 +42,18 @@ const devToolsConfig = {
...
 }
 
+const analyticsCaller = () => store => next => action => {
+  if (window.analytics) return next(action)
+
+  window.analytics = {
+    track(key, properties) {
+      store.dispatch(actions.analytics.trackEvent({ key, properties }))
+    }
+  }
```
dsisto-bc, line 52: Well played Mr Dylan, well played

blockdylanb: Hah! Yeah so I needed a way to hijack `window.analytics` that also had access to dispatch so I could fire off our normal trackEvent actions. After a bunch of failed attempts at creating a new saga, I thought maybe a redux middleware could work and I think this is actually a really nice solution, albeit a hacky one.

packages/blockchain-wallet-v4-frontend/src/modals/Onboarding/CompleteProfile/index.tsx
```diff
@@ -107,6 +108,11 @@ class CompleteProfile extends PureComponent<Props, State> {
     this.props.buySellActions.fetchAccumulatedTrades({ product: ProductTypes.SIMPLEBUY })
 
     const { currentStep } = this.props.data
+    peeksheetViewed({
+      current_step_completed: String(currentStep),
+      device: '',
+      platform: ''
```
blockdylanb, line 114: This is unfortunate. Since we require `device` and `platform` on our segment events typescript enforces these. A possible solution for this is to make these optional in segment, knowing that we will always add them to the events anyways.

milan-bc: we have those as fixed strings
```
device: 'WEB',
platform: 'WALLET'
```
might do the very same.

packages/blockchain-wallet-v4-frontend/src/store/index.js
milan-bc: amazing one 🔥

milan-bc (approved):
looks neat!!! Should we then remove all recent analytics work and migrate stuff to typewrite logic?

package.json
```diff
@@ -211,6 +212,7 @@
     "ts-loader": "8.3.0",
     "typecov": "0.2.3",
     "typescript": "4.5.5",
+    "typewriter": "^7.4.1",
```
schnogz, line 215: ping the versions and rerun yarn to update lockfile

packages/blockchain-wallet-v4-frontend/src/store/index.js
schnogz: can you add comments explaining what you're assigning/overriding on window? 
this looks like a middleware. should it be in that folder?

packages/blockchain-wallet-v4-frontend/src/modals/Onboarding/CompleteProfile/index.tsx
schnogz: possible to have some sort of wrapper that auto-adds these or maybe modify the typewriting script to make these optional to the types?

schnogz (changes requested):

packages/blockchain-wallet-v4-frontend/src/modals/Onboarding/CompleteProfile/index.tsx
blockdylanb: @mitchmalinin and I were discussing how to do this today. My initial thought is that we somehow `import *` all the typewriter methods, do something like `type NewPeekSheetViewed = Omit<PeekSheetViewed, 'device' | 'platform'>` and then re-export them (or a wrapper function that calls them) with this new type so we can keep the all the other type goodness from typewriter. device and platform will get added to the actual event through Milan's `trackEvent` work before being sent up to our api.

packages/blockchain-wallet-v4-frontend/src/store/index.js
blockdylanb: Moved this to it's own file and added comments.

PR #4207, opened 2022-02-15
fix(Swap/CoinSelection): display the correct selected account

awiermanncasas-bc:
The coin selection page was using only the account name to check
if it is the select account, but that is a problem when multiple
accounts have the same name. This improves the selection check by
doing a deep comparison on the account to check for the selected
account

## After fix:
https://user-images.githubusercontent.com/99212903/154145161-05daa4c5-65fb-4f1a-870a-2a4e4115144f.mov

### DIFF
--- a/packages/blockchain-wallet-v4-frontend/src/modals/Swap/CoinSelection/index.tsx
+++ b/packages/blockchain-wallet-v4-frontend/src/modals/Swap/CoinSelection/index.tsx
@@ -1,6 +1,7 @@
 import React, { PureComponent } from 'react'
 import { FormattedMessage } from 'react-intl'
 import { connect, ConnectedProps } from 'react-redux'
+import { equals } from 'ramda'
 
 import { Icon, Text } from 'blockchain-info-components'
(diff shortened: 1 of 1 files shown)


packages/blockchain-wallet-v4-frontend/src/modals/Swap/CoinSelection/index.tsx
```diff
@@ -1,6 +1,7 @@
 import React, { PureComponent } from 'react'
 import { FormattedMessage } from 'react-intl'
 import { connect, ConnectedProps } from 'react-redux'
+import { isEqual } from 'lodash'
```
blockdylanb, line 4: we don't have the lodash module installed on wallet. We use a library called Ramda that has a similar method though https://ramdajs.com/docs/#equals

awiermanncasas-bc: ok, will change that to use ramdajs.

blockdylanb (approved):
Looks good!

PR #4210, opened 2022-02-16
fix: removing unneeded component and using theme colors for styling

cmonti-bc:
## Description (optional)
Addressing previous PR comments.

### DIFF
--- a/packages/blockchain-wallet-v4-frontend/src/scenes/Login/components/UrlNoticeBar/index.tsx
+++ b/packages/blockchain-wallet-v4-frontend/src/scenes/Login/components/UrlNoticeBar/index.tsx
@@ -5,14 +5,11 @@ import styled from 'styled-components'
 import { Icon, Link, Text } from 'blockchain-info-components'
 import { media } from 'services/styles'
 
-const BACKGROUND_COLOR = '#0C6CF2'
-const FONT_COLOR = 'white'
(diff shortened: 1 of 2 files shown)


packages/blockchain-wallet-v4-frontend/src/scenes/Login/components/UrlNoticeBar/index.tsx
```diff
@@ -5,14 +5,11 @@ import styled from 'styled-components'
 import { Icon, Link, Text } from 'blockchain-info-components'
 import { media } from 'services/styles'
 
-const BACKGROUND_COLOR = '#0C6CF2'
-const FONT_COLOR = 'white'
 const BC_URL = 'https://login.blockchain.com'
```
pedroapfilho, line 8: this should come from the config

pedroapfilho (changes requested):

packages/blockchain-wallet-v4-frontend/src/scenes/Login/components/UrlNoticeBar/index.tsx
cmonti-bc: Sorry pedro but I couldn't find any example, if you do please let me know. For now I'll move forward according to andrews suggestion. Thanks again.

image

pedroapfilho (approved):

PR #4214, opened 2022-02-16
fix(Flayout): remove bottom flayout space

awiermanncasas-bc:
This remove the bottom flayout space by making
the flayout have a height of 100vh and removing
the bottom padding

## Testing Steps
* open any flayout in a mobile device or a mobile constrains

### DIFF
--- a/packages/blockchain-wallet-v4-frontend/src/components/Flyout/index.tsx
+++ b/packages/blockchain-wallet-v4-frontend/src/components/Flyout/index.tsx
@@ -73,8 +73,7 @@ const FlyoutModal = styled(AnimatedModal)`
   color: ${(props) => props.theme.grey700};
   ${media.mobile`
     width: 100%;
-    height: calc(100vh - 80px);
-    padding-bottom: 20px;
+    height: 100vh;
   `};
 `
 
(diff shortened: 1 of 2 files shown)


packages/blockchain-wallet-v4-frontend/src/components/Flyout/index.tsx
```diff
@@ -73,8 +73,7 @@ const FlyoutModal = styled(AnimatedModal)`
   color: ${(props) => props.theme.grey700};
   ${media.mobile`
     width: 100%;
-    height: calc(100vh - 80px);
-    padding-bottom: 20px;
+    height: 100vh;
```
blockdylanb, line 76: On production iOS Mobile Safari there is a banner that I believe is what that 80px is accounting for. I think this change could possibly push the flyout down an additional 80px more than it should and could cover up the buy button. You may want to use something other than `vh` in this case.
image

blockdylanb (changes requested):
See inline comment

packages/blockchain-wallet-v4-frontend/src/components/Flyout/index.tsx
awiermanncasas-bc: I was not able to open the page with my device
should someone with an iPhone try to debug this?

PR #4222, opened 2022-02-21
feat(CircularImage): adds a circular image component to components

awiermanncasas-bc:
This adds a new component to render circular images
like avatars, badges and symbols
Screen Shot 2022-02-21 at 5 01 06 PM


ethan-bc:
@awiermanncasas-bc 
can this accept SVGs as well? we're trying to use svgs as much as possible moving forward, and support for svgs would be dope.

awiermanncasas-bc:
> @awiermanncasas-bc can this accept SVGs as well? we're trying to use svgs as much as possible moving forward, and support for svgs would be dope.

What we can do is, drop this PR and add the coins svg logos to the constellation, then we just use the Icon component from constellation to display the coin logo.
Then all coins logo will be in svg and reusable from other projects

What you think about that @ethan-bc ?

PR #4224, opened 2022-02-22
add coin page scaffold

awiermanncasas-bc:
This PR adds the coin page component and the inner layout of the page
with the page header displaying the coin name, code, logo and description
Screen Shot 2022-02-22 at 12 29 16 PM

### DIFF
--- a/packages/blockchain-wallet-v4-frontend/src/pages/CoinPage/components/CoinHeader/CoinHeader.tsx
+++ b/packages/blockchain-wallet-v4-frontend/src/pages/CoinPage/components/CoinHeader/CoinHeader.tsx
@@ -1,0 +1,30 @@
+import React from 'react'
+
+import { Image, Text } from 'blockchain-info-components'
+
+import { Flex } from '../../../../components/Flex'
+import { CoinHeaderComponent } from './types'
+
+export const CoinHeader: CoinHeaderComponent = ({ coinCode, coinDescription, coinName }) => {
+  return (
+    <Flex gap={16}>
+      <Image name='btc' style={{ height: '48px', width: '48px' }} />
+
+      <Flex flexDirection='column' gap={4} justifyContent='center'>
+        <Flex gap={4}>
+          <Text color='grey900' weight={600} size='20px'>
+            {coinName}
+          </Text>
+
+          <Text color='grey600' weight={600} size='20px'>
+            {coinCode}
+          </Text>
+        </Flex>
+
+        <Text color='grey900' weight={500} size='16px'>
+          {coinDescription}
+        </Text>
+      </Flex>
+    </Flex>
+  )
+}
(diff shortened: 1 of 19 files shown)


packages/blockchain-wallet-v4-frontend/src/pages/CoinPage/components/PageScaffold/styles.ts
```diff
@@ -0,0 +1,15 @@
...
+  justify-content: center;
+`
+
+export const Grid = styled.div`
+  display: grid;
+  grid-template-columns: 2fr 1fr;
+  column-gap: 32px;
+  padding: 40px 64px;
+  width: 100%;
+  max-width: 1280px;
```
awiermanncasas-bc, line 14: @blockdylanb  do we have a specification for the max width of a page?
or @ethan-bc should we have this specification in constellation?

packages/blockchain-wallet-v4-frontend/src/pages/CoinPage/components/CoinHeader/CoinHeader.tsx
```diff
@@ -0,0 +1,30 @@
+import React from 'react'
+
+import { Image, Text } from 'blockchain-info-components'
```
awiermanncasas-bc, line 3: @blockdylanb should we use components from the `blockchain-info-components` or should we implement these in constellation too?

blockdylanb: Images still come from info components, so I think what you have here is ok for now. Text could possibly come from constellation but I haven't actually used it yet so would need to see if it's working properly.

packages/blockchain-wallet-v4-frontend/src/pages/CoinPage/components/PageScaffold/styles.ts
blockdylanb: Ethan mentioned having a max width in our meeting but I don't think he had decided on an actual number at that point. Let's go with 1280 for now and adjust as needed. 👍🏻

blockdylanb (approved):

packages/blockchain-wallet-v4-frontend/src/pages/CoinPage/components/CoinHeader/CoinHeader.tsx
awiermanncasas-bc: yes, I created a spike PR for the text in constellation but I think we still need to wait for Pedro to have the theme done and then they will revisit the Text component in constellation.

So I think there is a long way until we can get the Text component from constellalation.

Should I leave like this for now?

What I could do is try to revisit  the Text component in the blockchain info component to make it type safe, what you think?

blockdylanb: @awiermanncasas-bc Given that it'll go away eventually I think you can just leave it as is for now.

PR #4259, opened 2022-03-01
chore(constellation): upgrade from 0.0.27 to 0.0.28

awiermanncasas-bc:
This upgrades constellation version and fix all the
incompatibilities

* This also changes the swap icon in the trade view because the icon it was used before dont exist anymore
* Sets the default font family to Inter so the nav bar icons do use the font
* Adds the old implementation of the Switch but with correct colors
* Implements the Copy button in the mobile view

## Before:
Screen Shot 2022-03-01 at 7 09 58 PM

Screen Shot 2022-03-01 at 6 48 21 PM
Screen Shot 2022-03-01 at 6 48 12 PM

## After:
Screen Shot 2022-03-01 at 6 35 43 PM
Screen Shot 2022-03-01 at 6 35 06 PM
Screen Shot 2022-03-01 at 7 28 45 PM

### DIFF
--- a/packages/blockchain-wallet-v4-frontend/package.json
+++ b/packages/blockchain-wallet-v4-frontend/package.json
@@ -74,7 +74,7 @@
   "dependencies": {
     "@babel/polyfill": "7.12.1",
     "@blockchain-com/components": "6.1.1",
-    "@blockchain-com/constellation": "0.0.27",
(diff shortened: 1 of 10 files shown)


blockdylanb (approved):
LGTM

PR #4275, opened 2022-03-03
feat: removing autofocus for email and password

cmonti-bc:
## Description (optional)
Removing login inputs autofocus to avoid validation on focus out when switching products.

nV5EpWlB8D


amadeo-bc:
Can we somehow autofocus but don't show the error state by default?

PR #4291, opened 2022-03-07
feat(Tabs): add tabs

awiermanncasas-bc:
Screen Shot 2022-03-07 at 4 31 52 PM
Screen Shot 2022-03-07 at 4 31 56 PM

### DIFF
--- a/packages/blockchain-wallet-v4-frontend/src/components/Tabs/Tabs.stories.tsx
+++ b/packages/blockchain-wallet-v4-frontend/src/components/Tabs/Tabs.stories.tsx
@@ -1,0 +1,52 @@
+import React, { useState } from 'react'
+import { ComponentMeta } from '@storybook/react'
+
+import { Tab, Tabs, TabsComponent } from '.'
+
+const tabsStoriesMeta: ComponentMeta<TabsComponent> = {
+  component: Tabs,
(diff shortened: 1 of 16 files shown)


packages/blockchain-wallet-v4-frontend/src/components/Tabs/Tabs.stories.tsx
```diff
@@ -0,0 +1,52 @@
...
+  >('live')
+
+  return (
+    <Tabs>
+      <Tab
+        badgeColor='green'
+        onClick={() => setSelectedTab('live')}
+        selected={selectedTab === 'live'}
+      >
+        Live
```
blockdylanb, line 26: We'll need to convert these to formatted messages for translations but it's fine like this for now.

blockdylanb (approved):

packages/blockchain-wallet-v4-frontend/src/components/Tabs/Tabs.stories.tsx
awiermanncasas-bc: ah yes, when we actually use this components we can do this
```
<Tabs>
  <Tab>
    <ForemttedMessage />
  </Tab>
</Tabs>
```

PR #4319, opened 2022-03-11
Feat/ci 62/debit card module validation for eligible users

julianromero-bc:
## Description
Added validation for showing module only to eligible users


packages/blockchain-wallet-v4-frontend/src/data/components/debitCard/sagas.ts
```diff
@@ -11,8 +11,8 @@ export default ({ api }: { api: APIType }) => {
       const data = yield call(api.getDCProducts)
       yield put(A.getProductsSuccess(data))
     } catch (e) {
-      const error = errorHandler(e)
-      yield put(A.getProductsFailure(error))
+      console.error('Failed to get card products', errorHandler(e))
```
milan-bc, line 14: any reason to keep console here?

schnogz (approved):
looks good other than the console.log mentioned by Milan

packages/blockchain-wallet-v4-frontend/src/data/components/debitCard/sagas.ts
julianromero-bc: I didn't want to miss this information and, in this flow, there is no message shown to the user. However, it might be wrong to do this. I couldn't find a clear idea in the project on how this things should be managed

julianromero-bc:
> looks good other than the console.log mentioned by Milan

Thanks, I just answered that but not sure which should be the correct way

packages/blockchain-wallet-v4-frontend/src/data/components/debitCard/sagas.ts
julianromero-bc: After DM with Milan decided to leave it

PR #4343, opened 2022-03-16
Fix/fwn 1120/tb recive copy bch

tbuchann:
## Description (optional)
Remove `bitcoincash:` from address and copy address
Screen Shot 2022-03-16 at 9 29 05 AM

## Testing Steps (optional)
Open receive modal, make sure when you click the copy address icon the copied address does not include `bitcoincash:`


schnogz (changes requested):
I think this functionality already exists in BCH utils (`packages/blockchain-wallet-v4/src/utils/bch.js`).  if not, please add and export a new util function to achieve this.

schnogz (approved):

PR #4354, opened 2022-03-16
refactor(CoinView): move chart to hook

awiermanncasas-bc:


packages/blockchain-wallet-v4-frontend/src/data/hooks/useSelector/useSelector.tsx
```diff
@@ -0,0 +1,9 @@
+import { useSelector as useReactReduxSelector } from 'react-redux'
+
+import { RootState } from 'data/rootReducer'
+
+type Selector<T> = (state: RootState) => T
+
+export const useSelector = <T extends unknown = unknown>(selector: Selector<T>): T => {
```
blockdylanb, line 7: Is this somehow different then the `useSelector` straight from react-redux?

plondon: Also, why not use mapStateToProps?

awiermanncasas-bc: I implemented this useSelector over the react-redux hook because of the types.
by default the useSelector component from react-redux was not using our RootState as the state type and it was not providing the correct type in the result.

but I will double check if I can complete remove this component and use the react-redux

@plondon I want to use the `useSelector` hook so multiple hooks in the CoinView can connect to the store and not  depend all in one mapStateToProps at the root of the component

PR #4379, opened 2022-03-21
Style/FWN-1071/add iban code format

lsena-bc:
## Description
Added IBAN code format (4 characters per block)
Screen Shot 2022-03-21 at 15 51 41


lsena-bc:
Wrong branch name convention. Used "style" instead "feat".

PR #4397, opened 2022-03-23
FWN-1375 - fix: adding 'notranslate' class to avoid seed words being translated

cmonti-bc:
## Description (optional)
Adding `className='notranslate'` attribute to the parent wrappers to avoid google translate services to translate the wallet's "magic words" (or "seed words"). Better description inside [the ticket](https://blockchain.atlassian.net/browse/FWN-1375).

### DIFF
--- a/packages/blockchain-wallet-v4-frontend/src/modals/Settings/RecoveryPhrase/ShowRecoveryWords/template.tsx
+++ b/packages/blockchain-wallet-v4-frontend/src/modals/Settings/RecoveryPhrase/ShowRecoveryWords/template.tsx
@@ -67,7 +67,7 @@ const WordsList = ({ handleBackArrow, handleNextButton, step, words }) => {
         words.map((word, index) => {
           if (index < 6) {
             return (
-              <WordBox>
+              <WordBox className='notranslate'>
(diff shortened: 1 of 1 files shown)


pedroapfilho:
Can you link the docs for this `notranslate` magic here?

cmonti-bc:
> Can you link the docs for this `notranslate` magic here?

Here’s the [translate-attribute spec](https://html.spec.whatwg.org/multipage/dom.html#the-translate-attribute).

And also [the browser compatibility](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/translate#browser_compatibility). Looks like Firefox and Internet Explorer doesn’t support this attribute.

image

It seems that react doesn’t support the `translate=”no”` attribute but it does support a `className=”notranslate”` as[ found in their own blog](https://developers.google.com/search/blog/2008/10/helping-you-break-language-barrier), and in some other [non-official places](https://stackoverflow.com/a/9629628), which makes the fix super easy, localized and convenient.

If `className=”notranslate”` is used on a parent element all of their child elements will also inherit this attribute which makes it very practical.

schnogz (approved):

PR #4398, opened 2022-03-23
feat(institutional-portal): use redirect from url param

TheLeoB:
## Description (optional)
Uses redirect from url for institutional logins. 

## Testing Steps (optional)

### DIFF
--- a/packages/blockchain-wallet-v4-frontend/src/data/auth/sagas.ts
+++ b/packages/blockchain-wallet-v4-frontend/src/data/auth/sagas.ts
@@ -78,9 +78,8 @@ export default ({ api, coreSagas, networks }) => {
 
   const exchangeLogin = function* (action) {
     const { code, password, username } = action.payload
-    const { redirect, userType } = yield select(selectors.auth.getProductAuthMetadata)
+    const { platform, redirect, userType } = yield select(selectors.auth.getProductAuthMetadata)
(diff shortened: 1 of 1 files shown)


packages/blockchain-wallet-v4-frontend/src/data/auth/sagas.ts
```diff
@@ -122,12 +122,8 @@ export default ({ api, coreSagas, networks }) => {
...
           break
         // web - institutional exchange login
-        case userType === 'institutional' && institutionalPortalEnabled:
-          window.open(
-            `${institutionalDomain}/institutional/portal/?jwt=${jwtToken}`,
-            '_self',
-            'noreferrer'
-          )
+        case userType === 'institutional' && institutionalPortalEnabled && redirect:
+          window.open(`${redirect}?jwt=${jwtToken}`, '_self', 'noreferrer')
```
schnogz, line 126: what if there is no redirect? should we leave the fallback?

TheLeoB: [this ticket](https://blockchain.atlassian.net/browse/FWN-1345) says that only users coming from .com/institutional should be directed to the institutional app after login. All other users who just click the footer ‘Institutional User’ in the login page should just go to the regular exchange app. So if there's no redirect, it should go to the default case and open the regular exchange app. I can add a comment to make it clear.

schnogz (approved):

PR #4419, opened 2022-03-24
fix(FWN-1331): remove next buttons and replace with appropriate cta

tbuchann:
## Description (optional)
Change CTA buttons: Buy Now/ Withdraw Now
Screen Shot 2022-03-24 at 2 19 58 PM

## Testing Steps (optional)

### DIFF
--- a/packages/blockchain-wallet-v4-frontend/src/components/Flyout/Brokerage/EnterAmount.tsx
+++ b/packages/blockchain-wallet-v4-frontend/src/components/Flyout/Brokerage/EnterAmount.tsx
@@ -354,26 +354,28 @@ const Account = ({
   )
 }
 
-const NextButton = ({ invalid, orderType, paymentAccount, pristine, submitting }) => {
-  return (
-    <Button
-      data-e2e={orderType === BrokerageOrderType.DEPOSIT ? 'submitDepositAmount' : 'withdrawNext'}
-      height='48px'
-      size='16px'
-      nature='primary'
-      type='submit'
-      fullwidth
-      disabled={invalid || pristine || submitting || !paymentAccount}
-      onClick={() => {}}
-    >
-      {submitting ? (
-        <HeartbeatLoader height='16px' width='16px' color='white' />
-      ) : (
-        <FormattedMessage id='buttons.next' defaultMessage='Next' />
-      )}
-    </Button>
-  )
-}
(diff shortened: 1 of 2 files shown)


packages/blockchain-wallet-v4-frontend/src/components/Flyout/Brokerage/EnterAmount.tsx
```diff
@@ -354,26 +354,28 @@ const Account = ({
...
+const PreviewButton = ({ invalid, orderType, paymentAccount, pristine, submitting }) => (
+  <Button
+    data-e2e={orderType === BrokerageOrderType.DEPOSIT ? 'submitDepositAmount' : 'withdrawNext'}
+    height='48px'
+    size='16px'
+    nature='primary'
+    type='submit'
+    fullwidth
+    disabled={invalid || pristine || submitting || !paymentAccount}
+    onClick={() => {}}
```
blockdylanb, line 366: I see that this wasn't on the old button either, but is this an issue? Does the form still submit even if there is no onclick handler?

packages/blockchain-wallet-v4-frontend/src/modals/BuySell/CheckoutConfirm/template.success.tsx
```diff
@@ -201,17 +201,6 @@ const Success: React.FC<InjectedFormProps<{ form: string }, Props> & Props> = (p
     props.buySellActions.cancelOrder(props.order)
   }
 
-  const paymentPartnerButton =
-    paymentPartner === BankPartners.YAPILY ? (
-      <FormattedMessage id='copy.next' defaultMessage='Next' />
```
blockdylanb, line 206: Are you sure it's ok to remove this text? When you make a buy with yapily you don't go to the order confirmation screen after this screen. You go to an approval screen, and then on to the order confirmation screen.

packages/blockchain-wallet-v4-frontend/src/components/Flyout/Brokerage/EnterAmount.tsx
tbuchann: its on the old button.

blockdylanb: Ok yeah I think since it's `type='submit'` the form just submits when the user clicks it so this works.

PR #4429, opened 2022-03-25
feat(buy flow): fixed issue with redirect if there are pending orders

milan-bc:
## Description (optional)
Fixed issue with redirect if there are pending orders

## Testing Steps (optional)


schnogz (dismissed):

blockdylanb (approved):
This code definitely needs a larger refactor so we can sort out all of this business logic, but this looks correct for now.

PR #4434, opened 2022-03-25
FWN-1370 instrument maintenance errors on segment

cmonti-bc:
## Description (optional)
I had to add a manual call to the `/event/publish` endpoint for tracking events on segment as the issue that happened was redux store not being instanced. Without that regular library for using segment wouldn't work as the library is coupled to the store.

## Testing Steps (optional)
To test this fix the user needs to check Segment debug page in the right env that the crash occurs.

The user should be able to see something like the following

### Wallet Options not loading
image

### Wallet currencies not loading
image

### DIFF
--- a/packages/blockchain-wallet-v4-frontend/src/index.js
+++ b/packages/blockchain-wallet-v4-frontend/src/index.js
@@ -6,6 +6,8 @@ import ReactDOM from 'react-dom'
 import { BrowserRouter } from 'react-router-dom'
 import styled from 'styled-components'
 
+import { analyticsTrackingNoStore } from 'utils/helpers'
+
 import {
   FontGlobalStyles,
   IconGlobalStyles,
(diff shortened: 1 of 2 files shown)


packages/blockchain-wallet-v4-frontend/src/index.js
```diff
@@ -81,6 +83,29 @@ configureStore()
...
+          name: 'Wallet Load Failure',
+          originalTimestamp: new Date(),
+          properties: {
+            error_message: getErrorMessage(e.toString())
+          }
+        }
+      ]
+    }
+    analyticsTrackingNoStore(data)
+    console.error(e)
```
cmonti-bc, line 110: changed the `console.info` to `console.error` per @schnogz request.

packages/blockchain-wallet-v4-frontend/src/utils/helpers.ts
```diff
@@ -74,3 +76,31 @@ export const toBase64 = (file: File): Promise<string> => {
     reader.onerror = (error) => reject(error)
   })
 }
+
+/*
+  Analytics usage without redux store coupling.
+*/
+export const analyticsTrackingNoStore = async (data) => {
```
cmonti-bc, line 83: Not quite sure we want this over here and also if we want this just for the issue I'm fixing, if that's the case then I would encapsulate all of the implementation configuration into this file also to avoid untidy code on the `index.js`

cmonti-bc:
> let connect offline

This is done!

schnogz (approved):

PR #4446, opened 2022-03-28
Fwn 1389 when in any of the sso login screens change the title of the page to blockchain com instead of blockchain com wallet

cmonti-bc:
## Description (optional)
Changing page titles, we still need @amadeo-bc confirmation.

8ACIDmLFvj

### DIFF
--- a/packages/blockchain-wallet-v4-frontend/src/index.html
+++ b/packages/blockchain-wallet-v4-frontend/src/index.html
@@ -3,27 +3,47 @@
   <head>
     <meta charset="UTF-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=1" />
-    <meta name="apple-itunes-app" content="app-id=493253309">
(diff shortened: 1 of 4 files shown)


packages/blockchain-wallet-v4-frontend/src/index.html
```diff
@@ -3,27 +3,47 @@
...
+    <script
+      defer
+      nonce="**CSP_NONCE**"
+      src="https://www.google.com/recaptcha/enterprise.js?render=**RECAPTCHA_KEY**"
+    ></script>
+    <script
+      defer
+      nonce="**CSP_NONCE**"
+      src="https://www.googletagmanager.com/gtm.js?id=GTM-KK99TPJ"
+    ></script>
```
schnogz, line 46: i think we should undo all changes to this file. we still want the full title here for SEO. its okay if the page title flickers imo

schnogz (changes requested):

PR #4483, opened 2022-04-04
feat: add filter for cards terminated in debit card module

julianromero-bc:
## Description 
Implement filter when obtaining cards for not retrieving TERMINATED ones

### DIFF
--- a/packages/blockchain-wallet-v4-frontend/src/data/components/debitCard/sagas.ts
+++ b/packages/blockchain-wallet-v4-frontend/src/data/components/debitCard/sagas.ts
@@ -18,9 +18,13 @@ export default ({ api }: { api: APIType }) => {
     }
   }
 
+  const filterTerminatedCards = (cards) => cards.filter((card) => card.status !== 'TERMINATED')
+
   const getCards = function* () {
     try {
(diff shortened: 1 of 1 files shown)


packages/blockchain-wallet-v4-frontend/src/data/components/debitCard/sagas.ts
```diff
@@ -18,9 +18,13 @@ export default ({ api }: { api: APIType }) => {
     }
   }
 
+  const filterTerminatedCards = (cards) => cards.filter((card) => card.status !== 'TERMINATED')
```
schnogz, line 21: how many statuses are there going to be?  we should create an enum sooner rather than later. dont want to litter code with same strings everywhere

schnogz (approved):

packages/blockchain-wallet-v4-frontend/src/data/components/debitCard/sagas.ts
julianromero-bc: I understand, but this has close to nothing defined. No it will be only be used here, in the moment I need to use this string in other place I will create a enum an avoid duplication

PR #4489, opened 2022-04-05
feat(google-pay): adding google pay production key

pedroapfilho:


packages/blockchain-wallet-v4-frontend/src/data/components/buySell/sagas.ts
```diff
@@ -151,10 +151,13 @@ export default ({ api, coreSagas, networks }: { api: APIType; coreSagas: any; ne
     return token
   }
 
-  const generateGooglePayToken = async (paymentRequest: google.payments.api.PaymentDataRequest) => {
+  const generateGooglePayToken = async (
+    paymentRequest: google.payments.api.PaymentDataRequest
+  ): Promise<string> => {
+    const environment = window.location.host === 'login.blockchain.com' ? 'PRODUCTION' : 'TEST'
```
schnogz, line 157: please get the env from wallet-options selector

schnogz (changes requested):

packages/blockchain-wallet-v4-frontend/src/data/components/buySell/sagas.ts
pedroapfilho: This is because we actually can only show this one if the host is that one, that I've validated on google.

It can't be used on pre-prod.

Should I still use the selector?

schnogz (approved):

PR #4501, opened 2022-04-07
feat(useRemote): add hook to access remote data from redux

awiermanncasas-bc:
## Description
Add useRemote hook


blockdylanb (approved):
🤯

PR #4511, opened 2022-04-12
refactor(useRemote): add useRemote to hooks from coin page

awiermanncasas-bc:
This refactor changed the hooks that access redux in the coin page to use the `useRemote` hook


packages/blockchain-wallet-v4-frontend/src/scenes/CoinPage/components/CoinPage/hooks/useRecurringBuyPanel/useRecurringBuyPanel.tsx
```diff
@@ -41,14 +41,14 @@ export const useRecurringBuyPanel: UseRecurringBuyPanelHook = ({ coin }) => {
   }, [openRecurringBuyBanner, trackOnClickLearnMore])
 
   const recurringBuysNode: ReactNode = useMemo(() => {
-    if (!recurringBuyListItems) return null
+    if (isLoading) return null
```
pedroapfilho, line 44: don't we have a loading state?

pedroapfilho (approved):

packages/blockchain-wallet-v4-frontend/src/scenes/CoinPage/components/CoinPage/hooks/useRecurringBuyPanel/useRecurringBuyPanel.tsx
blockdylanb: We do not. If there are no recurring buys we show an advertisement for recurring buys, but if they're loading we don't currently have a loading state for them. We could add one but it's not in the designs yet as well as there is minimal layout shift when they do load.

blockdylanb (approved):

PR #4544, opened 2022-04-18
fix(coin view): get coin from router for legacy transactions component

blockdylanb:
## Description (optional)

## Testing Steps (optional)

### DIFF
--- a/packages/blockchain-wallet-v4-frontend/src/scenes/Transactions/index.tsx
+++ b/packages/blockchain-wallet-v4-frontend/src/scenes/Transactions/index.tsx
@@ -1,6 +1,7 @@
 import React from 'react'
 import { FormattedMessage } from 'react-intl'
 import { connect, ConnectedProps } from 'react-redux'
+import { RouteComponentProps } from 'react-router'
 import { path, toLower } from 'ramda'
 import { bindActionCreators, compose, Dispatch } from 'redux'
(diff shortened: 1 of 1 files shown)


packages/blockchain-wallet-v4-frontend/src/scenes/Transactions/index.tsx
```diff
@@ -308,10 +310,15 @@ class TransactionsContainer extends React.PureComponent<Props> {
 }
 
 const mapStateToProps = (state, ownProps: OwnProps): LinkStatePropsType =>
-  getData(state, ownProps.coin, ownProps.coinfig)
+  getData(
+    state,
+    ownProps.computedMatch.params.coin,
+    window.coins[ownProps.computedMatch.params.coin].coinfig
```
plondon, line 316: Potentially unsafe

plondon (approved):

packages/blockchain-wallet-v4-frontend/src/scenes/Transactions/index.tsx
blockdylanb: Previously `window.coins` and `coin` props were passed in from app.js, but were still essentially doing this same thing. I could pass in window.coins from app.js but not the coin because we get it from react-router now. I could also get the coinfig in selectors.ts and add some checks to see if it exists or else throw.

PR #4545, opened 2022-04-19
Add alias to hook

awiermanncasas-bc:
this adds a new alias to the hooks folder and update the current files that use the hooks to use the new alias


schnogz (changes requested):
please alphabetize the sorting of the alias registrations. makes them easiest to find.

PR #4546, opened 2022-04-19
Feat/ci 99/add terminate card feature

julianromero-bc:
## Description 
Add Terminate Card feature

## DEMO
demo-terminate-card

## NOTES
It is in pending status the dashboard loading state, which will be shown when opening the dashboard and also when terminating a card


packages/blockchain-wallet-v4-frontend/src/data/components/debitCard/sagas.ts
```diff
@@ -69,9 +70,21 @@ export default ({ api }: { api: APIType }) => {
...
 
+  const terminateCard = function* (action: ReturnType<typeof A.terminateCard>) {
+    try {
+      const { payload } = action
+      yield call(api.terminateDC, payload)
+
+      yield call(getCards)
+    } catch (e) {
+      console.error('Failed to terminate card', errorHandler(e))
+    }
```
schnogz, line 81: this feels like it needs to be a Remote.  how are you going to inform the user that we failed to terminate the card?

schnogz (changes requested):

packages/blockchain-wallet-v4-frontend/src/data/components/debitCard/sagas.ts
julianromero-bc: We Are still discussing how to show that to the user with Cody. Also, this is for MVP, in V1 designs this feature will be shown in a different way. So I prefer not to spend much time on this

schnogz (approved):

PR #4549, opened 2022-04-20
fix: typo after login on /home tab

cmonti-bc:
## Description (optional)

FcFeVTsebe


schnogz (changes requested):
Lets just keep this simple because your solution does not work for all use cases (sometimes users dont go to the homepage after login). Just change `layouts/Wallet/index.tsx` to set the page title to `Blockchain.com Wallet`.

cmonti-bc:
> Lets just keep this simple because your solution does not work for all use cases (sometimes users dont go to the homepage after login). Just change `layouts/Wallet/index.tsx` to set the page title to `Blockchain.com Wallet`.

schnogz (approved):

devildp333:
ขอบคุณจากหัวใจ

PR #4609, opened 2022-04-27
feat: add error handling for lock card feature

julianromero-bc:
## Description 
Add message for lock card feature when success or error

## Demo (success)
demo-lock-card-message

## Demo (error)
demo-lock-card-error-message


packages/blockchain-wallet-v4-frontend/src/components/Alerts/messages.tsx
```diff
@@ -182,6 +182,20 @@ const getAlertContent = (message, data = undefined) => {
           defaultMessage='Currency has been successfully updated'
         />
       )
+    case C.DEBIT_CARD_DASHBOARD_LOCK_ERROR:
+      return buildMessageTemplate(
+        <FormattedMessage
+          id='components.alerts.debit_card_dashboard_lock_error'
+          defaultMessage='Failed to change Card Lock'
```
blockdylanb, line 189: does "Card Lock" need to be capitalized?

packages/blockchain-wallet-v4-frontend/src/scenes/DebitCard/CardDashboard/ManageCardBox/index.tsx
```diff
@@ -24,10 +25,25 @@ const SwitchWrapper = styled.div`
   margin: auto;
 `
 
-const ManageCardBox = ({ cards, debitCardActions, lockHandler, modalActions }) => {
+const ManageCardBox = ({ alertActions, cards, debitCardActions, lockHandler, modalActions }) => {
   const activeCard = cards[0]
   const [isCardLocked, setCardLocked] = useState(activeCard.status === CardStateType.LOCKED)
 
+  useEffect(() => {
+    lockHandler.cata({
```
blockdylanb, line 33: I think this could end up having weird side effects. Try using the new `useRemote()` hook that @awiermanncasas-bc created. Outside of the useEffect you would have something 
```
 const { hasData, hasError } = useRemote(lockHandler)

useEffect(() => {
  if (hasData && !hasError) {
    alertActions.displaySuccess(C.DEBIT_CARD_DASHBOARD_LOCK_SUCCESS)
  } else if (!hasData && hasError) {
    setCardLocked(!isCardLocked)
    alertActions.displayError(C.DEBIT_CARD_DASHBOARD_LOCK_ERROR)
  }
}, [hasData, hasError])
```

packages/blockchain-wallet-v4-frontend/src/scenes/DebitCard/CardDashboard/index.tsx
```diff
@@ -11,6 +11,7 @@ const Iframe = styled.iframe`
 `
 
 const CardDashboard = ({
+  alertActions,
```
blockdylanb, line 14: Instead of passing actions as a param to child components just do it here in this component. The index.tsx files are more for the getting/setting data and the template files are for displaying the data. 
You could do:
```
const alertErrorCallback = useCallback(() => {
  alertActions.displayError(C.DEBIT_CARD_DASHBOARD_LOCK_ERROR)
}, [])

...
<ManageCardBox
  alertErrorHandler={alertErrorCallback}
  ...
/>
```

packages/blockchain-wallet-v4-frontend/src/components/Alerts/messages.tsx
julianromero-bc: Discussions with the UX about these messages are still pending

packages/blockchain-wallet-v4-frontend/src/scenes/DebitCard/CardDashboard/ManageCardBox/index.tsx
blockdylanb: I was thinking particularly about if `lockHandler` were to change too often it would cause the useEffect to rerender the whole component, but after looking at this again, even in my example if lockHandler were to change the component would rerender so maybe they are functionally the same. Stylistically though with useRemote, you don't have to include the loading and NotAsked states since they don't do anything anyways. I guess it's up to you to decide which way you like best.

julianromero-bc: I prefer this way because I think it is easier to read and understand what it does in each state. In the other case you mentioned, you have to read 2 different if clauses and try to understand which state of the request it is handleling

blockdylanb (approved):

PR #4651, opened 2022-05-03
fix/CI-127/debit-card-module-view

julianromero-bc:
## Description
removed hardcoded boolean value and add logic for showing debit card if user is eligible
fix module initialization

## DEMO
demo-debit-card-view


plondon (approved):
LGTM! plus nice cleanup on the wallet layout container !

PR #4698, opened 2022-05-06
feat(captcha): consolidate captcha token fetch into saga

schnogz:
## Description
- consolidates captcha logic in a single place and refreshes the token right before needed by the API to hopefully fix obscure bugs


packages/blockchain-wallet-v4/src/network/api/profile/index.ts
```diff
@@ -14,9 +14,8 @@ export default ({ authorizedGet, authorizedPost, authorizedPut, get, nabuUrl, po
       endPoint: '/signin',
       url: nabuUrl
     })
-  }
 
-  const exchangeResetPassword = (email) => {
+  const exchangeResetPassword = (email) =>
```
schnogz, line 18: @TheLeoB was missing a return

TheLeoB (approved):

PR #4739, opened 2022-05-12
chore(react-query): add react-query and add the in memory cache for queries

awiermanncasas-bc:
## Description
Add react-query as an alternative to handle asynchronous states


julianromero-bc:
Could you explain in which case is this being used? I saw the` app.tsx` but is mostly an ident change after adding the the wrapper

pedroapfilho (approved):

milan-bc (approved):
:shipit:

PR #4866, opened 2022-05-25
feat(brokerage): move more api errors to better track in segment and …

blockdylanb:
…refactor class component

## Description (optional)

## Testing Steps (optional)


schnogz (approved):
We really need to migrate to the new events model.  then we dont have actions that are so confusing they require a comment to explain what they are for.

PR #4870, opened 2022-05-26
feat: adding ExpansionPanel component in bank details for Argentinian users, and also removing ARS currency from the default currencies (this is only for Argentina)

lsena-bc:
## Description
Adding show/hide functionality to the Argentinian bank details, and also the ARS currency was removed from the default currencies because this feature is only for Argentina.

Screen Shot 2022-05-26 at 13 29 40

### DIFF
--- a/packages/blockchain-wallet-v4-frontend/src/components/ExpansionPanel/ExpansionPanel.tsx
+++ b/packages/blockchain-wallet-v4-frontend/src/components/ExpansionPanel/ExpansionPanel.tsx
@@ -1,4 +1,4 @@
-import React, { useState } from 'react'
+import React, { useCallback, useState } from 'react'
 import { FormattedMessage } from 'react-intl'
 import { Icon } from '@blockchain-com/constellation'
 import { IconChevronDown, IconChevronUp } from '@blockchain-com/icons'
@@ -10,7 +10,11 @@ import { Padding } from 'components/Padding'
 import { ExpansionPanelComponent } from './types'
 
 export const ExpansionPanel: ExpansionPanelComponent = ({ children }) => {
-  const [isToggled, handleToggle] = useState(false)
+  const [isToggled, setToggle] = useState(false)
+
+  const handleToggle = useCallback(() => {
+    setToggle((t) => !t)
+  }, [isToggled])
 
   return (
     <>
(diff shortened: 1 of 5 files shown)


packages/blockchain-wallet-v4-frontend/src/components/ExpansionPanel/ExpansionPanel.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+  return (
+    <div>
+      {isToggled ? <div>{children}</div> : ''}
+      <Button
+        nature='white-blue'
+        small
+        style={{ margin: '8px auto 0px' }}
+        data-e2e={isToggled ? 'showLess' : 'showMore'}
+        // @ts-ignore
```
blockdylanb, line 22: What is the ts error this is ignoring?

packages/blockchain-wallet-v4-frontend/src/components/ExpansionPanel/ExpansionPanel.tsx
```diff
@@ -0,0 +1,41 @@
...
+export const ExpansionPanel: ExpansionPanelComponent = ({ children }) => {
+  const [isToggled, handleToggle] = useState(false)
+
+  return (
+    <div>
+      {isToggled ? <div>{children}</div> : ''}
+      <Button
+        nature='white-blue'
+        small
+        style={{ margin: '8px auto 0px' }}
```
blockdylanb, line 20: Could you remove this style prop and wrap the button in a `<Padding></Padding>` component? Here's some instructions https://blockchain.github.io/blockchain-wallet-v4-frontend/?path=/story/structural-padding-intro--page

lsena-bc: Thank you Dylan!! Could you re check it please?

lsena-bc:
> i would use more destructing on the props as well as create variables for some as they are accessed multiple times. this way you avoid the super long and repeated object prop drilling
> 
> for example, make a variable out of the account address since its used 5+ times and that way you only have to cast it once
> 
> `const accountAddress = props.account.agent as AgentType).address`

Done, thank you for the feedback @schnogz !

packages/blockchain-wallet-v4-frontend/src/components/ExpansionPanel/ExpansionPanel.tsx
```diff
@@ -0,0 +1,41 @@
...
+  const [isToggled, handleToggle] = useState(false)
+
+  return (
+    <>
+      {isToggled ? <div>{children}</div> : ''}
+      <Padding top={10}>
+        <Button
+          nature='white-blue'
+          small
+          style={{ margin: '8px auto 0px' }}
```
blockdylanb, line 22: can remove this line now since you added the padding?

blockdylanb (dismissed):

schnogz (approved):

PR #4896, opened 2022-05-31
feat(ExternalSvg): add component to display svg from url

awiermanncasas-bc:
## Description
This adds the ExternalSVG component to be able to render an svg from a url.
This uses the image tag inside a svg to be able to fetch and render the svg, also handle failure and provides a fallback attribute

### example
```
<ExternalSVG
  src="/svg-path.svg"
  width={1} // in rem
  height={1} // in rem
  fallback={(
    <FallbackImageOrComponent />
  )}
/>
```
Screen Shot 2022-05-31 at 2 44 46 PM
Screen Shot 2022-05-31 at 2 44 54 PM

### DIFF
--- a/packages/blockchain-wallet-v4-frontend/src/components/ExternalSVG/ExternalSVG.test.tsx
+++ b/packages/blockchain-wallet-v4-frontend/src/components/ExternalSVG/ExternalSVG.test.tsx
@@ -1,0 +1,43 @@
+import React from 'react'
+import { shallow } from 'enzyme'
+
+import { ExternalSVG } from '.'
+
+describe('ExternalSVG', () => {
+  it('Should display the svg image', () => {
+    const wrapper = shallow(<ExternalSVG src='/assets/icon.svg' width={3} height={2} />)
+
+    expect(wrapper.type()).toEqual('svg')
+    expect(wrapper.getElement().props.width).toEqual('3rem')
+    expect(wrapper.getElement().props.height).toEqual('2rem')
+
+    const image = wrapper.find('image')
+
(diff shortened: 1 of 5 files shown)


packages/blockchain-wallet-v4-frontend/src/components/ExternalSVG/ExternalSVG.test.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+    const imageProps = image.getElement().props
+
+    expect(wrapper.find('[data-testid="image-fallback"]').exists()).toBe(false)
+
+    imageProps.onError()
+
+    expect(wrapper.find('[data-testid="image-fallback"]').exists()).toBe(true)
+  })
+})
```
julianromero-bc, line 43: Nice tests! 🤩

packages/blockchain-wallet-v4-frontend/src/components/ExternalSVG/index.ts
```diff
@@ -0,0 +1,2 @@
+export { default as ExternalSVG } from './ExternalSVG'
```
julianromero-bc, line 1: I don't think you need to add the alias in the export

packages/blockchain-wallet-v4-frontend/src/components/ExternalSVG/ExternalSVG.test.tsx
milan-bc: we need more tests like this imo :)

milan-bc (approved):

PR #4956, opened 2022-06-07
fix(order): removing flexible pricing order

pedroapfilho:


schnogz (approved):

julianromero-bc (approved):
It also nice to see that useless code is being removed

PR #4976, opened 2022-06-09
feat(Extra KYC): default context string changed

fgonzalez-bc:
Last minute change, DEFAULT_CONTEXT string to DEFAULT


julianromero-bc (approved):
Thats why I always ask to remove hardcoded strings xD

PR #4991, opened 2022-06-14
feat(signup): load list of countries from API

milan-bc:
## Description (optional)
This PR fix signup to use API and load list of available countries from there instead of using hardcoded list

## Testing Steps (optional)


packages/blockchain-wallet-v4-frontend/src/data/auth/sagas.ts
```diff
@@ -61,7 +61,9 @@ export default ({ api, coreSagas, networks }) => {
   const LOGIN_FORM = 'login'
 
   const authNabu = function* () {
-    yield put(actions.components.identityVerification.fetchSupportedCountries())
+    yield put(
+      actions.components.identityVerification.fetchSupportedCountries({ scope: CountryScope.KYC })
+    )
```
schnogz, line 66: why do we need to fetch country list here, is there a better spot?  this function is for authentication to nabu, seems like breaking single responsibility principles

packages/blockchain-wallet-v4-frontend/src/scenes/Signup/components/SignupForm/index.tsx
```diff
@@ -66,9 +65,37 @@ const PasswordRequirementText = styled(Text)<{ isValid?: boolean }>`
...
+
+const getStateElements = (states: Array<StateType>) => [
+  {
+    group: '',
+    items: states.map((state: StateType) => ({
+      text: state.name,
+      value: state
+    }))
+  }
+]
```
schnogz, line 86: are these always US states?  if so, maybe rename things to indicated that? `usStates`?

packages/blockchain-wallet-v4-frontend/src/scenes/Signup/types.ts
```diff
@@ -16,10 +16,10 @@ export type GeoLocationType = {
   ip: string
 }
 export type SignupFormType = {
-  country: string
+  country: CountryType
   email: string
   password: string
-  state: string
+  state: StateType
```
schnogz, line 22: `usState`?

schnogz (changes requested):
Can we refactor further to completely remove the `SelectBoxUSState` and `SelectBoxCountry` components and data?  Seems like there is only one other spot using them in BuySell flows

packages/blockchain-wallet-v4-frontend/src/scenes/Signup/index.tsx
```diff
@@ -153,6 +157,10 @@ const mapStateToProps = (state: RootState): LinkStatePropsType => ({
   isLoadingR: selectors.signup.getRegistering(state) as RemoteDataType<string, undefined>,
   language: selectors.preferences.getLanguage(state),
   search: selectors.router.getSearch(state) as string,
+  supportedCountries: selectors.components.identityVerification
+    .getSupportedCountries(state)
+    .getOrElse([]),
```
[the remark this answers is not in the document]
milan-bc: we can't I was just about to render page itself and not to show error, but yeah you are right

packages/blockchain-wallet-v4-frontend/src/scenes/Signup/components/SignupForm/index.tsx
milan-bc: they are, yeah we can rename but BE expect to get just state will do convert when sending it back to BE

packages/blockchain-wallet-v4-frontend/src/data/components/identityVerification/types.ts
```diff
@@ -47,6 +47,8 @@ export type ScopesType = Array<'KYC' | 'Mercury'>
 export type CampaignsType = 'BLOCKSTACK' | 'POWER_PAX' | 'SUNRIVER'
 
 export type StateType = {
+  code: string
+  countryCode: string
```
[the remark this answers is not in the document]
milan-bc: for example:

```
code: "US-AL"
countryCode: "US"
```

packages/blockchain-wallet-v4-frontend/src/data/auth/sagas.ts
milan-bc: I have to check with @TheLeoB she added this in the first place, here I'm just changing call signature since we now pass scope parameter.

schnogz (approved):

PR #5023, opened 2022-06-23
fix(login): adding props to FormError so the design is not broken

pedroapfilho:
BEFORE

Screen Shot 2022-06-23 at 4 34 17 PM

AFTER

Screen Shot 2022-06-23 at 4 34 02 PM


julianromero-bc:
> BEFORE
> 
> Screen Shot 2022-06-23 at 4 34 17 PM
> 
> AFTER
> 
> Screen Shot 2022-06-23 at 4 34 02 PM

is it expected to have removed the red border on the error in input?

julianromero-bc (approved):
You have my approve if it is expected that change

PR #5041, opened 2022-06-27
fix: select input "ValueContainer" clickable area

cmonti-bc:
## Description (optional)

nMB7In2nbk

### DIFF
--- a/packages/blockchain-info-components/src/Form/SelectInput/template.js
+++ b/packages/blockchain-info-components/src/Form/SelectInput/template.js
@@ -5,6 +5,8 @@ import styled, { css } from 'styled-components'
 
 import { selectBorderColor, selectFocusBorderColor } from '../helper'
 
+const INPUT_ID = 'clickableInput'
+
 // 🚨🚨🚨
 // react-select overrides
 // shared with CreatableInput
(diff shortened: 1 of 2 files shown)


packages/blockchain-info-components/src/Form/SelectInput/template.js
```diff
@@ -125,6 +125,11 @@ export const sharedSelect = css`
 
 const StyledSelect = styled(Select)`
   ${sharedSelect}
+
+  .dummy-input-wrapper {
+    position: absolute;
+    top: -1000px;
+  }
```
cmonti-bc, line 132: So react-select always adds a `DummyInput` component to the `ValueContainer` but we weren't rendering it so the events that should be passed down/binded to the container where never there. This is for hiding that `DummyInput`.

packages/blockchain-info-components/src/Form/SelectInput/template.js
```diff
@@ -227,6 +238,7 @@ const SelectInput = (props) => {
           Option,
           ValueContainer
         }}
+        inputId={INPUT_ID}
```
cmonti-bc, line 241: image

packages/blockchain-info-components/src/Form/SelectInput/template.js
julianromero-bc: Did you try just using `visibility: hidden;` ?

julianromero-bc (approved):
if it works I would prefer using the visibility css style

packages/blockchain-info-components/src/Form/SelectInput/template.js
cmonti-bc: https://github.com/JedWatson/react-select/issues/2597

plondon (commented):
Any issue w/ other dropdowns? Some to check:

WalletBalanceDropdown on coin view
account dropdown on send crypto screens (legacy, BTC, ETH, XLM, BCH)

PR #5086, opened 2022-07-01
feat: disable "Add [coin]" button when coin not eligible on interest flow

cmonti-bc:
## Description (optional)

On the left we have a user with eligible coins and on the right we have a US user with no eligibility but with already used rewards account so it's possible to buy or to withdraw the coins

IIA080tmDp


blockdylanb (approved):

schnogz (changes requested):

schnogz:
backmerging the release into this branch. no need for this pr

PR #5099, opened 2022-07-05
feat: disable "Add [coin]" button when coin not eligible on interest flow

cmonti-bc:
On the left we have a user with eligible coins and on the right we have a US user with no eligibility but with already used rewards account so it's possible to buy or to withdraw the coins.

176955691-162a5004-9f54-48a3-b40f-5cd62a982b68

### DIFF
--- a/packages/blockchain-wallet-v4-frontend/src/modals/Interest/AccountSummary/template.success.tsx
+++ b/packages/blockchain-wallet-v4-frontend/src/modals/Interest/AccountSummary/template.success.tsx
@@ -63,15 +63,7 @@ const AccountSummary: React.FC<Props> = (props) => {
   const interestBalanceStandard = convertBaseToStandard(coin, interestBalanceBase)
   const pendingInterestStandard = convertBaseToStandard(coin, pendingInterestBase)
 
-  const getInterestEligibility = () => {
-    const interestEligibility = interestEligible[coin]
-    if (!interestEligibility) {
-      return false
-    }
-    return interestEligibility.eligible
-  }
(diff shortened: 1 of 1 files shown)


packages/blockchain-wallet-v4-frontend/src/modals/Interest/AccountSummary/template.success.tsx
```diff
@@ -62,6 +63,16 @@ const AccountSummary: React.FC<Props> = (props) => {
...
 
+  const getInterestEligibility = () => {
+    const interestEligibility = interestEligible[coin]
+    if (!interestEligibility) {
+      return false
+    }
+    return interestEligibility.eligible
+  }
+
+  const isDepositEnabled = getInterestEligibility()
```
schnogz, line 74: this isnt very performant. defining a function on every render and then invoking it. why not just the following, i believe it is the same result

`const isDepositEnabled = interestEligible[coin] ? interestEligible[coin]?.eligible : false`

schnogz (changes requested):

packages/blockchain-wallet-v4-frontend/src/modals/Interest/AccountSummary/template.success.tsx
cmonti-bc: Sure, that also does it, thanks andrew. I was worried about the different structure of the response when it's eligible and when it's not but that does the trick. Will change it.

schnogz (approved):

PR #5110, opened 2022-07-06
fix(order): fixing race condition for enter amount card polling

pedroapfilho:
Fixes rare race condition, where order is still on polling, and we try to cancel and recreate another one on the background


awiermanncasas-bc (approved):
👍

blockdylanb (approved):

PR #5153, opened 2022-07-13
feat(amount): adding feature to save last unused amount on cache

pedroapfilho:


packages/blockchain-wallet-v4-frontend/src/data/components/buySell/sagas.ts
```diff
@@ -401,6 +402,13 @@ export default ({ api, coreSagas, networks }: { api: APIType; coreSagas: any; ne
 
       yield put(A.createOrderLoading())
 
+      yield put(
+        cacheActions.setLastUnusedAmount({
+          amount: convertBaseToStandard('FIAT', input.amount),
```
julianromero-bc, line 407: There is a constant FIAT defined, not sure if the place where it was defined is correct but it would be great to use that instead of a hardcoded string

packages/blockchain-wallet-v4-frontend/src/data/components/buySell/sagas.ts
```diff
@@ -1536,9 +1554,13 @@ export default ({ api, coreSagas, networks }: { api: APIType; coreSagas: any; ne
...
         .getOrElse(false) as boolean
 
+      const lastUnusedAmounts = selectors.cache.getLastUnusedAmounts(yield select())
+
+      const lastUnusedAmount = lastUnusedAmounts ? lastUnusedAmounts[pair.pair] : null
+
       yield put(
         actions.form.initialize(FORM_BS_CHECKOUT, {
-          amount,
+          amount: amount || (fix === 'FIAT' && lastUnusedAmount) ? lastUnusedAmount : '0',
```
julianromero-bc, line 1563: duplicated 'FIAT' string

packages/blockchain-wallet-v4-frontend/src/data/components/buySell/sagas.ts
pedroapfilho: This is typed for a string, I think it should keep this way, as it can be `FIAT` or any other `coin`

packages/blockchain-wallet-v4-frontend/src/data/components/buySell/sagas.ts
pedroapfilho: same as above

PR #5174, opened 2022-07-15
feat(extra kyc): extra trigger in buy/sell

fgonzalez-bc:


packages/blockchain-wallet-v4-frontend/src/services/extraKYC/index.ts
```diff
@@ -0,0 +1,39 @@
+import { call, put, take } from 'redux-saga/effects'
+
+import { APIType } from '@core/network/api'
+import { ExtraKYCContext } from '@core/types'
+import { actions } from 'data'
+import { VerifyIdentityOriginType } from 'data/types'
+
+export const getExtraKYCCompletedStatus = function* ({
```
blockdylanb, line 8: It seems strange to have this method in the services folder instead of auth or signup sagas in `src/data`. Any particular reason you did it this way?

fgonzalez-bc: I don't think is part of auth or signup, it is a trigger for extra questions before differents flows starts. In fact is for t2 users.

lsena-bc: Done! @blockdylanb

PR #5197, opened 2022-07-20
fix(AppleAndGooglePay): this changes the banner order

awiermanncasas-bc:
Updates the order of the banners so apple and google pay
is first, and catch error when trying to check for apple pay
availability on safari

### DIFF
--- a/packages/blockchain-wallet-v4-frontend/src/hooks/useApplePay/utils/isApplePayAvailable/isApplePayAvailable.ts
+++ b/packages/blockchain-wallet-v4-frontend/src/hooks/useApplePay/utils/isApplePayAvailable/isApplePayAvailable.ts
@@ -1,2 +1,11 @@
-export const isApplePayAvailable = (): boolean =>
-  (window.ApplePaySession && ApplePaySession?.canMakePayments()) || false
+export const isApplePayAvailable = (): boolean => {
+  if (!window.ApplePaySession) return false
+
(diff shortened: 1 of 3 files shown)


packages/blockchain-wallet-v4-frontend/src/hooks/useApplePay/utils/isApplePayAvailable/isApplePayAvailable.ts
```diff
@@ -1,2 +1,11 @@
-export const isApplePayAvailable = (): boolean =>
-  (window.ApplePaySession && ApplePaySession?.canMakePayments()) || false
+export const isApplePayAvailable = (): boolean => {
+  if (!window.ApplePaySession) return false
+
+  try {
+    return ApplePaySession?.canMakePayments()
```
pedroapfilho, line 5: you don't need this question mark here, you already checked it on the top

pedroapfilho (changes requested):

packages/blockchain-wallet-v4-frontend/src/hooks/useApplePay/utils/isApplePayAvailable/isApplePayAvailable.ts
awiermanncasas-bc: ok, I removed the ?

PR #5200, opened 2022-07-21
feat(NabuError): log all nabu errors to segment

awiermanncasas-bc:
## Description
Add an interception to axios to intercept and log all nabu errors to segment

## Testing Steps
- Run app locally and open the [segment debugger](https://app.segment.com/blockchain/sources/web_staging/debugger) for staging
- Now for every nabu error you should see the error in segment a few seconds later

Screen Shot 2022-07-21 at 11 38 33 AM
Screen Shot 2022-07-21 at 11 38 52 AM

### DIFF
--- a/packages/blockchain-wallet-v4-frontend/src/services/errors/NabuError/utils/createNabuErrorAnalyticsInterceptor/createNabuErrorAnalyticsInterceptor.ts
+++ b/packages/blockchain-wallet-v4-frontend/src/services/errors/NabuError/utils/createNabuErrorAnalyticsInterceptor/createNabuErrorAnalyticsInterceptor.ts
@@ -1,0 +1,29 @@
+import { actions } from 'data'
+import { ClientErrorProperties } from 'data/analytics/types/errors'
+import { Analytics } from 'data/types'
+
+import { isNabuError } from '../isNabuError'
+import { CreateNabuErrorAnalyticsInterceptorUtility } from './createNabuErrorAnalyticsInterceptor.types'
+
+export const createNabuErrorAnalyticsInterceptor: CreateNabuErrorAnalyticsInterceptorUtility =
+  (dispatch) => (error) => {
+    if (isNabuError(error)) {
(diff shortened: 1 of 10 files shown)


pedroapfilho (approved):

milan-bc (approved):

packages/blockchain-wallet-v4-frontend/src/services/errors/NabuError/utils/createNabuErrorAnalyticsInterceptor/createNabuErrorAnalyticsInterceptor.ts
```diff
@@ -0,0 +1,29 @@
...
+
+      dispatch(
+        actions.analytics.trackEvent({
+          key: Analytics.CLIENT_ERROR,
+          properties: clientError
+        })
+      )
+    }
+
+    return Promise.reject(error)
```
blockdylanb, line 28: Is this line what would normally happen if this interceptor didn't exist and the api returned a 4xx or 5xx? So if the error is not a nabu error, it won't effect any of our existing error handling?

pedroapfilho: exactly

blockdylanb (approved):

packages/blockchain-wallet-v4-frontend/src/services/errors/NabuError/utils/createNabuErrorAnalyticsInterceptor/createNabuErrorAnalyticsInterceptor.ts
awiermanncasas-bc: yes, it will not affect the current handling because the error will always go through and the normal handlers should take care of the error

PR #5230, opened 2022-07-26
fix(order): fixing skip error display list

pedroapfilho:
This is already fixed on development, but not here.


blockdylanb (approved):

awiermanncasas-bc (commented):
👍

PR #5274, opened 2022-08-05
feat(2fa): adding required 5 digits for 2 factor auth

pedroapfilho:


packages/blockchain-wallet-v4-frontend/src/scenes/Login/Exchange/TwoFA/index.tsx
```diff
@@ -91,7 +91,7 @@ const TwoFAExchange = (props: Props) => {
             nature='primary'
             fullwidth
             height='48px'
-            disabled={submitting || invalid || busy || !formValues?.exchangeTwoFA}
+            disabled={submitting || invalid || busy || (formValues?.exchangeTwoFA?.length || 0) < 5}
```
blockdylanb, line 94: shouldn't a 2FA length be 6?

TheLeoB: This page is only for legacy exchange login which only uses auth app, so let's bump to 6. Wallet/TwoFA should be 5.

TheLeoB (approved):

TheLeoB (approved):

PR #5321, opened 2022-08-23
Adds Private and Trading account flyouts

awiermanncasas-bc:
## Description
Adds flyout for private key and trading account in coin view v2
Screen Shot 2022-08-23 at 2 59 45 PM
Screen Shot 2022-08-23 at 2 59 38 PM


packages/blockchain-wallet-v4-frontend/src/components/IconCircularBackground/IconCircularBackground.tsx
```diff
@@ -4,27 +4,47 @@ import { colors } from '@blockchain-com/constellation'
...
       style={{
-        alignItems: 'center',
-        backgroundColor: colors[color] ?? color,
-        borderRadius: '50%',
-        display: 'flex',
         height: sizeToPixels[size] ?? size,
-        justifyContent: 'center',
+        position: 'relative',
         width: sizeToPixels[size] ?? size
       }}
```
milan-bc, line 23: any reason why not creating a separate styled component?

packages/blockchain-wallet-v4-frontend/src/modals/ViewPrivateKeyWalletFlyout/ViewPrivateKeyWalletFlyout.tsx
```diff
@@ -0,0 +1,315 @@
...
+            {receiveListTile}
+            <Divider />
+            {sendListTile}
+            <Divider />
+            {swapListTile}
+            <Divider />
+            {sellListTile}
+            <Divider />
+          </>
+        )}
```
milan-bc, line 307: those 2 looks almost identical you can use check of hasFunds on line 286 and prevent duplications

packages/blockchain-wallet-v4-frontend/src/hooks/useOpenReceiveModal/useOpenReceiveModal.ts
```diff
@@ -0,0 +1,33 @@
...
+      await new Promise((r) => setTimeout(r, duration))
+
+      dispatch(actions.form.change(REQUEST_FORM, 'selectedAccount', wallet))
+
+      dispatch(actions.form.change(REQUEST_FORM, 'step', RequestSteps.SHOW_ADDRESS))
+    },
+    [dispatch]
+  )
+
+  return [openReceiveModal]
```
milan-bc, line 32: is there a some reason why this is in `[]` since here it's returning only one const maybe it can be just that one and then easier for import on places where it is used

packages/blockchain-wallet-v4-frontend/src/modals/ViewPrivateKeyWalletFlyout/ViewPrivateKeyWalletFlyout.tsx
awiermanncasas-bc: I see, I think the problem is that when the user have no funds, we have to show the list with a different order.
So when the user has funds we have to show the send button first and when they have no funds we have to show the receive button first.

Could you show an example of how the code could look like?

packages/blockchain-wallet-v4-frontend/src/hooks/useOpenReceiveModal/useOpenReceiveModal.ts
awiermanncasas-bc: yes, makes sense, will update to return the funtion

packages/blockchain-wallet-v4-frontend/src/modals/ViewPrivateKeyWalletFlyout/ViewPrivateKeyWalletFlyout.tsx
milan-bc: sure like this


```
        <>
            <Divider />
            {hasFunds ? sendListTile : receiveListTile}
            <Divider />
            {hasFunds ?  receiveListTile: sendListTile}
            <Divider />
            {swapListTile}
            <Divider />
            {sellListTile}
            <Divider />
          </>
```

milan-bc (approved):
LGTM

PR #5332, opened 2022-08-30
feat(referral landing): added support for referral landing goals for cowboys

milan-bc:
## Description (optional)
This PR add only support for deep linking and running goals for referral landing of cowboys campaign 

## Testing Steps (optional)


packages/blockchain-wallet-v4-frontend/src/data/goals/sagas.ts
```diff
@@ -144,6 +144,19 @@ export default ({ api, coreSagas, networks }) => {
     )
   }
 
+  const defineCowboysCampaignGoal = function* (search) {
```
blockdylanb, line 147: Actually, I think we just keep this goal more generic and not cowboys related. We'll be able to use this to deep link to referrals for a lot of different use-cases.

packages/blockchain-wallet-v4-frontend/src/data/goals/sagas.ts
```diff
@@ -144,6 +144,19 @@ export default ({ api, coreSagas, networks }) => {
...
+    // /#/open/cowboys?token_id=456
```
blockdylanb, line 148: How about something like `/#/open/referrals` which will open the referral flyout. Then in the `useEffect()` we check for the user tag `COWBOYS_2022` to display the proper branding?

milan-bc: yeah I was unsure how we want to have this from marketing perspective

pedroapfilho (approved):

PR #5387, opened 2022-09-25
fix(create-user): create user if credentials empty in metadata

TheLeoB:
## Description (optional)
Addresses issue reported by Peter - we only call `/users` to create or get a nabu user if it's the users first login. This is an issue for older wallets (created pre-nabu) since login throws an error like this. 
Screen Shot 2022-09-24 at 12 28 28 PM (1) (1)

Main fix is we pass `firstLogin` boolean from loginRoutineSaga into nabu profile signIn function. If it's false but nabu credentials are empty, we call the createUser saga. 

## Testing Steps (optional)

### DIFF
--- a/packages/blockchain-wallet-v4-frontend/src/data/modules/profile/sagas.ts
+++ b/packages/blockchain-wallet-v4-frontend/src/data/modules/profile/sagas.ts
@@ -220,54 +220,6 @@ export default ({ api, coreSagas, networks }) => {
     yield call(setSession, nabuUserId, nabuLifetimeToken, email, guid)
   }
 
-  const signIn = function* () {
-    try {
-      const email = (yield select(selectors.core.settings.getEmail)).getOrFail('No email')
-      const guid = yield select(selectors.core.wallet.getGuid)
-      // TODO: in future only fetch unified credentials
-      const unifiedNabuCredentialsR = yield select(
-        selectors.core.kvStore.unifiedCredentials.getNabuCredentials
-      )
-      const userCredentialsR = yield select(
-        selectors.core.kvStore.userCredentials.getLegacyNabuCredentials
-      )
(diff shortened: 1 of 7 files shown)


schnogz (approved):
Looks good!

TheLeoB:
> Looks good! I was trying to remember how to not have to use the `ts-ignore`'s in the register file but I cant remember :(

same same. i asked milan and pedro in the slack chat if they remembered. I think the saga has to return a type

packages/blockchain-wallet-v4-frontend/src/data/modules/profile/sagas.ts
```diff
@@ -589,6 +541,58 @@ export default ({ api, coreSagas, networks }) => {
...
+      )).getOrElse({})
+      if (!unifiedNabuCredentials.nabuUserId || !unifiedNabuCredentials.nabuLifetimeToken) {
+        yield put(
+          actions.core.kvStore.unifiedCredentials.setUnifiedCredentials({
+            nabu_lifetime_token: nabuLifetimeToken,
+            nabu_user_id: nabuUserId
+          })
+        )
+      }
+      if (!firstLogin && (!nabuUserId || !nabuLifetimeToken)) {
```
TheLeoB, line 577: Main change is here, call `/users` if there's no nabuID in metadata. The error in the screenshot was being thrown at 581.

TheLeoB:
> > Looks good! I was trying to remember how to not have to use the `ts-ignore`'s in the register file but I cant remember :(
> 
> same same. i asked milan and pedro in the slack chat if they remembered. I think the saga has to return a type

AHH, i remembered. Fixed it.

milan-bc (approved):
🚀

milan-bc (approved):

PR #5425, opened 2022-10-11
fix(buySell): revert add recurring buy option in checkout page

dkremniov-bc:
## Description
This PR reverts commit 2045eab9 which has introduced a bug with recurring buys using BANK_TRANSFER payment method.


milan-bc:
@dkremniov-bc great work so far, your title and. commit message are not in align with standards we do use

blockdylanb:
This looks good to me, I'm going to hold off on approving/merging though. 4.73 was delayed until tomorrow so let's hold off of on merging this until it does to prod. Then we can merge this and follow up with a hot deploy so we can test it independently and not potentially delay the release any more.

blockdylanb (approved):

PR #5446, opened 2022-10-19
fix(recurringBuy): filter out inactive recurring buys from response

dkremniov-bc:
## Description
Due to Back-End change, inactive recurring orders are returned in the recurring orders list response. Right now, if user deletes a recurring buy order - it still appears on the page. The solution is to implement filtering on our side.

### DIFF
--- a/packages/blockchain-wallet-v4-frontend/src/data/components/recurringBuy/slice.test.ts
+++ b/packages/blockchain-wallet-v4-frontend/src/data/components/recurringBuy/slice.test.ts
@@ -1,0 +1,32 @@
+import Remote from '@core/remote'
+import {
+  RecurringBuyItemState,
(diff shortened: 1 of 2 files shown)


milan-bc (approved):

packages/blockchain-wallet-v4-frontend/src/data/components/recurringBuy/slice.test.ts
```diff
@@ -0,0 +1,32 @@
...
+      ).toEqual(
+        expect.objectContaining({
+          registeredList: Remote.of(activeBuys)
+        })
+      )
+    })
+  })
+
+  it.todo('other cases')
+})
```
milan-bc, line 32: 👏

PR #5449, opened 2022-10-20
feat(api): helper for using API service with axios

milan-bc:
## Description (optional)

## Testing Steps (optional)

### DIFF
--- a/packages/blockchain-wallet-v4-frontend/src/hooks/useUserAddress/useUserAddress.ts
+++ b/packages/blockchain-wallet-v4-frontend/src/hooks/useUserAddress/useUserAddress.ts
@@ -1,0 +1,39 @@
+import { useQuery } from 'react-query'
+import { useSelector } from 'react-redux'
+import axios from 'axios'
+
+import { selectors } from 'data'
+import { RootState } from 'data/rootReducer'
+
+import { UserAddressData, UserAddressHook } from './useUserAddress.types'
+
+const useUserAddress: UserAddressHook = (searchText: string, country?: string) => {
+  const {
+    data: { api }
+  } = useSelector(selectors.core.walletOptions.getDomains)
+
+  const nabuToken = useSelector((state: RootState) =>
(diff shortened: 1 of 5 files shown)


packages/blockchain-wallet-v4-frontend/src/hooks/useUserAddress/useUserAddress.ts
```diff
@@ -0,0 +1,39 @@
...
+import { useSelector } from 'react-redux'
+import axios from 'axios'
+
+import { selectors } from 'data'
+import { RootState } from 'data/rootReducer'
+
+import { UserAddressData, UserAddressHook } from './useUserAddress.types'
+
+const useUserAddress: UserAddressHook = (searchText: string, country?: string) => {
+  const {
```
blockdylanb, line 11: Something you could do here is make `ApiHelper` a hook. I would call it `useAuthQuery` or `useNabuAuthQuery`. It can decorate useQuery with the nabu endpoint and api token and return an instance of useQuery to be called by `useUserAddress`

Something like this:
```
const useUserAddress: UserAddressHook = (searchText: string, country?: string) => {
  const urlForAddresses = `/address-capture/find?text=${searchText}${countryParams}`
  const [get, /* post, put, delete */] = useNabuAuthQuery()

  return get(urlForAddresses, searchText)
}
```

mvashchuk-bc: My two cents: I would keep `useQuery` as it is imported from react-query directly where we use it and solve authentication before that on the axios instance level. So a fetch function that we pass to `useQuery` uses a correct instance depends on it's needs.

Two reasons:
1. To me it looks like it's responsibility of an instance not useQuery, which is responsible for the request status and what data/errors to handle.
2. We will most likely complicate our life creating copies of useQuery by having a lot types issues. react-query has pretty complex types especially if you passing options, use refetch or infinite queries. I would not get that problems to think about in the future.

PR #5462, opened 2022-10-25
fix(Home): fix the way how linear domain is padded

dkremniov-bc:
## Description
Application of padding relative to the height of the chart was causing issue with scale for assets with lower price. Thus, the padding values should be relative to the values domain.

| Before  | After |
| ------------- | ------------- |
| Screenshot 2022-10-25 at 14 20 22 | Screenshot 2022-10-25 at 14 20 26  |
| Screenshot 2022-10-25 at 14 20 31  | Screenshot 2022-10-25 at 14 20 35  |


dkremniov-bc:
@pedroapfilho, I would appreciate your input. Any chance you remember what was [the intention here](https://github.com/blockchain/blockchain-wallet-v4-frontend/pull/3797/files#diff-fbc6d8f1a1b90ff67d02033fd23997a5eadd8adfa2d3215d4eddae80bd98f9e1R86): to add some padding to avoid chart being cut off or to make the size of the chart smaller? If the latter, then we should increase the padding in this PR.

PR #5525, opened 2022-11-15
feat(sardine): improved sardine flow to use risk api end point and check are flows on list

milan-bc:
## Description (optional)
This is a sardine improvement to use risk API endpoint and based on that set a flow

## Testing Steps (optional)

### DIFF
--- a/packages/blockchain-wallet-v4-frontend/src/data/modules/profile/selectors.ts
+++ b/packages/blockchain-wallet-v4-frontend/src/data/modules/profile/selectors.ts
@@ -169,9 +169,7 @@ export const isExchangeRelinkRequired = (state): RemoteDataType<string, boolean 
   })(getUserData(state))
 
 export const isFlowInRiskSettings = (state: RootState, flowName: string): boolean => {
-  const userRiskSettings = getUserRiskSettings(state).getOrElse({ flows: [] } as UserRiskSettings)
+  const userRiskSettings: UserRiskSettings = getUserRiskSettings(state).getOrElse({ flows: [] })
 
-  return !!(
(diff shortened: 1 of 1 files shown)


packages/blockchain-wallet-v4-frontend/src/data/modules/profile/selectors.ts
```diff
@@ -166,3 +167,11 @@ export const isExchangeRelinkRequired = (state): RemoteDataType<string, boolean
   lift((user: UserDataType) => {
     return not(isNil(prop('settings', user))) && length(getRemainingCoins(state))
   })(getUserData(state))
+
+export const isFlowInRiskSettings = (state: RootState, flowName: string): boolean => {
+  const userRiskSettings = getUserRiskSettings(state).getOrElse({ flows: [] } as UserRiskSettings)
```
dkremniov-bc, line 172:
```suggestion
  const userRiskSettings: UserRiskSettings = getUserRiskSettings(state).getOrElse({ flows: [] })
```

I would avoid type casting if possible.

packages/blockchain-wallet-v4-frontend/src/data/modules/profile/selectors.ts
```diff
@@ -166,3 +167,11 @@ export const isExchangeRelinkRequired = (state): RemoteDataType<string, boolean
...
+
+  return !!(
+    userRiskSettings.flows.length && userRiskSettings.flows.some((ur) => ur.name === flowName)
+  )
```
dkremniov-bc, line 176:
```suggestion
  return userRiskSettings.flows.some((ur) => ur.name === flowName)
```

Can be simplified.

dkremniov-bc (approved):

PR #5536, opened 2022-11-21
feat(dex-nav): layout and base navigation

mvashchuk-bc:
## Description (optional)
First version of constellation navigation to test inside DEX. Only layout and basic navigation is there. Still a lot to handle in the next PRs e.g. app QR code, user dropdown, 

## Testing Steps (optional)
Go to DEX and check it

### DIFF
--- a/packages/blockchain-wallet-v4-frontend/src/layouts/Dex/DexHeader/DexHeader.tsx
+++ b/packages/blockchain-wallet-v4-frontend/src/layouts/Dex/DexHeader/DexHeader.tsx
@@ -1,9 +1,9 @@
 import React from 'react'
 import { useIntl } from 'react-intl'
-import { useDispatch } from 'react-redux'
+import { useDispatch, useSelector } from 'react-redux'
 import { IconPhone, IconRefresh, Navigation } from '@blockchain-com/constellation'
 
-import { actions } from 'data'
+import { actions, selectors } from 'data'
 
 type Props = {
   selectedTab?: 'home' | 'prices' | 'earn' | 'nfts' | 'dex'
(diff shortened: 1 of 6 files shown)


packages/blockchain-wallet-v4-frontend/src/layouts/Dex/DexHeader/DexHeader.tsx
```diff
@@ -1,103 +1,207 @@
...
 
-const mapStateToProps = (state) => ({
-  ethBalanceR: selectors.balances.getCoinNonCustodialBalance('ETH')(state)
-})
-
-const connector = connect(mapStateToProps)
-type Props = ConnectedProps<typeof connector> & OwnProps
-
-export default connector(DexHeader)
+export const DexHeader = withRouter<Props>(Header)
```
blockdylanb, line 207: After our conversation I looked into this a little more. I think instead of wrapping the component in `withRouter` and using `history.push()` you can just dispatch a route change action like this: `dispatch(actions.router.push('/home'))` you can see a few examples in the code if you grep for that string.

mvashchuk-bc: Thanks, that works! Fixed

packages/blockchain-wallet-v4-frontend/src/layouts/Dex/DexHeader/DexHeader.tsx
```diff
@@ -1,103 +1,230 @@
...
+    }
+  ]
+
+  return (
+    <Navigation
+      defaultSelected={selectedTab}
+      iconActions={iconActions}
+      navigationTabs={navigationTabs}
+      dropdownSecondSectionItems={accountDropdownItems}
+      onSelectedChange={(key) => {
```
mperdomo-bc, line 158: I think this is pretty big for being inside as a prop, wouldn't it be better to have it outside?

In a pretty minor improvement, you could do 
```
case 'home':
case 'prices':
case 'earn':
case 'nfts':
case 'dex':
  dispatch(actions.router.push('/' + key))
  break
```

duckman2277 (approved):

duckman2277 (approved):

PR #5545, opened 2022-11-28
Refactor(dex-swap): refactor dex swap scenes

mvashchuk-bc:
## Description (optional)
Refactor of the existing layout and some logic there. Moved some components to use constellation where possible. It's pretty huge but really hard to decouple that efficiently. Next ones will be smaller and more atomic in terms of logic.


mvashchuk-bc:
Thanks for review @milan-bc! Fixed the comments

PR #5547, opened 2022-11-29
Feat/vgs integration

blockdylanb:
## Description (optional)
Adds a new credit card provider called VGS (verygoodsecurity.com) which replicates the add card flow that we also use checkout.com for and so the add card UI will look mostly the same as it does currently in production. There is also a new screen in the buy flow if we need the user to re-enter their CVV number before checking out they'll see the screenshot below.
image

## Testing Steps (optional)


packages/blockchain-wallet-v4-frontend/src/modals/BuySell/UpdateSecurityCode/UpdateSecurityCode.tsx
```diff
@@ -0,0 +1,171 @@
...
+  const updateCvv = () => {
+    const paymentId = order?.attributes?.paymentId
+    if (cvv && paymentId) {
+      dispatch(actions.components.buySell.updateCardCvv({ cvv, paymentId }))
+    }
+  }
+
+  useEffect(() => {
+    if (cvvHasData) {
+      setTimeout(() => {
```
dkremniov-bc, line 53: Do we absolutely need the timeout here and how reliable is it?

packages/blockchain-wallet-v4-frontend/src/modals/BuySell/AddCardVgs/AddCardVgs.tsx
```diff
@@ -0,0 +1,127 @@
...
+  const ref = useRef<HTMLIFrameElement>(null)
+
+  const { data, error, isLoading, isNotAsked } = useRemote(getData)
+  const { cardTokenId, vgsVaultId } = useSelector(selectors.components.buySell.getVgsAddCardInfo)
+  const cardSuccessRate = useSelector(selectors.components.buySell.getCardSuccessRate)
+  const formValues = useSelector(
+    selectors.form.getFormValues(FORMS_BS_BILLING_ADDRESS)
+  ) as BSBillingAddressFormValuesType
+
+  const handleReceivedPostMessage = useCallback(
```
dkremniov-bc, line 35: That would be a great candidate for refactoring into saga in my opinion 😀

packages/blockchain-wallet-v4-frontend/src/modals/BuySell/AddCardVgs/AddCardVgs.tsx
```diff
@@ -0,0 +1,127 @@
...
+        handleRetry={handleAction}
+        handleReset={handleAction}
+        handleBack={handleAction}
+      />
+    )
+  }
+  if (isPolling) return <CardLoading polling />
+  if (error) return null
+  if (!data || isLoading || isNotAsked) return <Loading text={LoadingTextEnum.GETTING_READY} />
+  if (!data) return null
```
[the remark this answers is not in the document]
mperdomo-bc: Depends on wether it should return null or the loading - because line 111 already takes `data` into account so line 112 would never be executed

packages/blockchain-wallet-v4-frontend/src/modals/BuySell/UpdateSecurityCode/UpdateSecurityCode.tsx
```diff
@@ -0,0 +1,171 @@
...
+                  postfix={
+                    cvvHasData ? (
+                      <IconCheckCircle color={SemanticColors.success} size='medium' />
+                    ) : (
+                      <IconLockClosed color={SemanticColors.muted} size='medium' />
+                    )
+                  }
+                  onChange={(e) => setCvv(e.currentTarget.value)}
+                  helperText={
+                    cvvHasError && (
```
mperdomo-bc, line 122: Question here - if `cvvHasError` is `null`/`undefined`/`false` it will return that value; can `helperText` parse those value correctly in the error cases?

packages/blockchain-wallet-v4-frontend/src/data/components/buySell/sagas.ts
```diff
@@ -595,8 +597,12 @@ export default ({ api, coreSagas, networks }: { api: APIType; coreSagas: any; ne
 
   const orderConfirmCheck = function* (orderId) {
     const order: ReturnType<typeof api.getBSOrder> = yield call(api.getBSOrder, orderId)
-
-    if (order.state === 'FINISHED' || order.state === 'FAILED' || order.state === 'CANCELED') {
+    if (
+      order.state === 'FINISHED' ||
+      order.state === 'FAILED' ||
+      order.state === 'CANCELED' ||
```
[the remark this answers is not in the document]
blockdylanb: I've gone back and fourth on converting everything to enums. In this case I think it's actually better to be a union type of strings because you get all the same benefits of an enum without the extra keystrokes. Also this PR would grow quite a bit if I made that change and it's already a pretty decent size. Checkout the thread I started about this in slack, and let's discuss.

packages/blockchain-wallet-v4-frontend/src/data/components/buySell/sagas.ts
```diff
@@ -2217,15 +2243,25 @@ export default ({ api, coreSagas, networks }: { api: APIType; coreSagas: any; ne
     }
   }
```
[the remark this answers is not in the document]
blockdylanb: Yeah totally. The issue is that since this call (and the products call) is in the saga, it makes a few seconds delay from when you click the button to when the flyout flies out. I think what we could do though is `dispatch()` these api calls from a `useEffect` in the component and set the response as a remote type in redux. So then the flyout will fly out immediately after clicking the button, but will sit on the loading state until all the api calls resolve.

packages/blockchain-wallet-v4-frontend/src/modals/BuySell/AddCardVgs/AddCardVgs.tsx
```diff
@@ -0,0 +1,127 @@
...
+  const handleAction = () => {
+    buySellActions.destroyCheckout()
+  }
+
+  useEffect(() => {
+    if (!cardSuccessRate) return
+
+    if (!ref.current?.contentWindow) {
+      return
+    }
```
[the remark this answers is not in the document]
blockdylanb: definitely.

packages/blockchain-wallet-v4-frontend/src/modals/BuySell/UpdateSecurityCode/UpdateSecurityCode.tsx
blockdylanb: I put this slight delay in order to show the user the green check in the postfix of the input as shown in the screenshot. Otherwise the once they submit the screen will just go to the next step with no confirmation to the user.
image

packages/blockchain-wallet-v4-frontend/src/modals/BuySell/UpdateSecurityCode/UpdateSecurityCode.tsx
blockdylanb: Fair point. It takes a react node so would be fine in case of null but perhaps not false. I'll be more explicit and pass `<></>`

packages/blockchain-wallet-v4-frontend/src/modals/BuySell/AddCardVgs/AddCardVgs.tsx
blockdylanb: After doing some more research here I think you're right. Using eventChannels https://redux-saga.js.org/docs/advanced/Channels/#using-the-eventchannel-factory-to-connect-to-external-events we could create a saga that listens for the postMessage events which would dispatch the other events that are already in this `handleReceivedPostMessage` function. I'm going to create a ticket for myself to do this, but don't want to hold up this merge as I need to get backend testing this in staging so I can sort out some CORS errors with the.

packages/blockchain-wallet-v4-frontend/src/modals/BuySell/UpdateSecurityCode/UpdateSecurityCode.tsx
dkremniov-bc: Makes sense.

By the way, even though it is a slight delay, you might want to clear the timeout in the cleanup function to avoid unexpected behaviour in edge cases.

packages/blockchain-wallet-v4-frontend/src/modals/BuySell/AddCardVgs/AddCardVgs.tsx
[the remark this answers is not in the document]
dkremniov-bc: That is actually a good example of why `switch` or `cata` is significantly more readable.

PR #5616, opened 2022-12-16
Feat/pay 553 deposit confirm ux

blockdylanb:
## Description (optional)
Adding 2 new rows to the buy confirm and deposit confirm screen indicating when the funds will be available to trade and withdraw.
image
image

## Testing Steps (optional)


packages/blockchain-wallet-v4-frontend/src/components/Brokerage/ConfirmAvailabilityRows/util.tsx
```diff
@@ -0,0 +1,104 @@
+import React from 'react'
```
dkremniov-bc, line 1: Personally, I didn't expect to see components in this file. I suggest calling this `ParseDepositTerms.tsx` instead of `util.tsx`.

packages/blockchain-wallet-v4-frontend/src/components/Brokerage/ConfirmAvailabilityRows/util.tsx
```diff
@@ -0,0 +1,104 @@
...
+import { FormattedMessage } from 'react-intl'
+import { add, format } from 'date-fns'
+
+import { DepositTerms, DisplayMode } from 'data/types'
+
+const Switcher = (
```
dkremniov-bc, line 7: This can be split into two subcomponents (e.g. `Title` and `Value`) to avoid variable reassignment.

packages/blockchain-wallet-v4-frontend/src/components/Brokerage/ConfirmAvailabilityRows/util.tsx
```diff
@@ -0,0 +1,104 @@
...
+  trade?: {
+    title: string | React.ReactNode
+    value: string | React.ReactNode
+  }
+  withdraw?: {
+    title: string | React.ReactNode
+    value: string | React.ReactNode
+  }
+}
+export const ParseDepositTerms = (depositTerms: DepositTerms): ParseDepositTermsReturnType => {
```
dkremniov-bc, line 93: Considering that this component is used in a single place, why did you opt out from just rendering `CheckoutRow` directly in this component? The way how this component is used feels weird.

packages/blockchain-wallet-v4-frontend/src/components/Brokerage/ConfirmAvailabilityRows/util.tsx
dkremniov-bc: To simplify this further, `Switcher` could be a hook just returning the text. Do we have a hook solution for internationalization?

packages/blockchain-wallet-v4-frontend/src/components/Brokerage/ConfirmAvailabilityRows/util.tsx
blockdylanb: yeah it didn't start out with components in it. I'll make that change.

dkremniov-bc:
Hooks look much better! It would also be great to unit test them.

dkremniov-bc (approved):

PR #5626, opened 2022-12-20
Lokalise: Translations update

blockchain-lokalize:
This pull request was initiated by Lokalise (user Lokalise User) at 2022-12-20 13:16:04


dsisto-bc:
I triggered this update from lokalise to update exchange promo keys. But seems to be too much here, closing for now.

PR #5631, opened 2022-12-20
Feat/active rewards

jjBlockchain:
## Description (optional)
* Add active rewards to activity page
* adjust coin and home view to account for active rewards balance


packages/blockchain-wallet-v4-frontend/src/data/components/interest/sagas.ts
```diff
@@ -400,16 +411,33 @@ export default ({ api, coreSagas, networks }: { api: APIType; coreSagas: any; ne
         })
       }
 
+      if (
```
blockdylanb, line 414: Would love to see some more comments in here as logic like this is pretty complex and will be hard for other developers to understand what's going on here.

blockdylanb (approved):

blockdylanb (approved):

PR #5673, opened 2023-01-09
reafactor(react router): updating react router package

blockdylanb:
## Update React Router
In order for us to fix many tech-debt related issues in the wallet (React 18, routing, loading states, # urls, etc.) we need to first update react router from v4 -> v6. This will be a difficult task but is the linch pin to getting the overall wallet application to a better state.

### Tasks
- [x] update react-router-dom package
- [ ] remove connected-react-router package
  - [ ] refactor any code using this package
- [ ] replace `<BrowserRouter />` with `createBrowserRouter()`
- [ ] refactor any use of `<Redirect />`. It's deprecated in favor of `redirect` util or `useNavigate` hook
- [ ] use the `useNavigate` or `useNavigation` hooks to navigate inside of components where we currently do in sagas
- [ ] Remove `<Switch />` component
- [ ] Instead of Layout components like `<AuthLayout />` and `<WalletLayout />` returning a `<Route />` component, it should look as much like a simple router as possible:
```
 <Routes>
  <Route path="/" element={<WalletLayout />}>
    <Route index element={<Home />} />
    <Route path="prices" element={<Prices />} />
    <Route path="*" element={<NoMatch />} />
  </Route>
</Routes>
```
We may also find it useful to implement our routing as a route object like so:
```
let routes: RouteObject[] = [{
  path: "/",
  element: <WalletLayout />,
  children: [
    { index: true, element: <Home /> },
    { path: "/prices", element: <Prices /> },
    { path: "*", element: <NoMatch /> },
  ]
}]
```
- [ ] ...

### DIFF
--- a/packages/blockchain-wallet-v4-frontend/package.json
+++ b/packages/blockchain-wallet-v4-frontend/package.json
@@ -141,7 +141,7 @@
     "react-qr-reader": "2.2.1",
     "react-redux": "7.2.3",
     "react-router-bootstrap": "0.25.0",
-    "react-router-dom": "4.3.1",
+    "react-router-dom": "6.6.1",
     "react-table": "7.8.0",
     "react-tooltip": "3.10.0",
     "react-use-measure": "2.0.4",
--- a/yarn.lock
+++ b/yarn.lock
@@ -5387,6 +5387,11 @@
     redux-thunk "^2.3.0"
     reselect "^4.0.0"
 
+"@remix-run/router@1.2.1":
+  version "1.2.1"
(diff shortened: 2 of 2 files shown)


dkremniov-bc:
Even though it is inevitable that we will need to remove usage of router in Redux, replacement of `connected-react-router` with [redux-first-history](https://github.com/salvoravida/redux-first-history) might be a reasonable first step to speed up this refactoring.

dkremniov-bc:
We shall probably start migration from upgrading to v5.1 as per these instructions: https://reactrouter.com/en/main/upgrading/v5#upgrade-to-react-router-v51

dkremniov-bc:
This will be addressed in separate PRs.

PR #5676, opened 2023-01-11
fix(dex-allowance): Layout (preparation before on chain interactions)

mvashchuk-bc:
It's just a layout with visibility logic of an allowance button. I'm preparing a mocked version of the whole flow before adding on chain interactions.

### DIFF
--- a/packages/blockchain-wallet-v4-frontend/src/scenes/Dex/Swap/AllowanceCheck/AllowanceCheck.tsx
+++ b/packages/blockchain-wallet-v4-frontend/src/scenes/Dex/Swap/AllowanceCheck/AllowanceCheck.tsx
@@ -1,0 +1,102 @@
+import React, { useState } from 'react'
+import { FormattedMessage, useIntl } from 'react-intl'
+import {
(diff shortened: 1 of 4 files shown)


packages/blockchain-wallet-v4-frontend/src/scenes/Dex/Swap/AllowanceCheck/AllowanceCheck.tsx
```diff
@@ -0,0 +1,102 @@
...
+const GlobalStyleTooltipZIndexFix = createGlobalStyle`
+  div[data-radix-popper-content-wrapper] {
+    z-index: 99999 !important;
+    width: 286px !important;
+    min-width: 286px !important;
+  }
+`
+
+// TODO: Update constellation Button to accept text color
+const GreenButtonWrapper = styled.div`
```
mvashchuk-bc, line 22: Not able to set it directly on Button because it just reset all the styles. So as a temp solution I had to add a wrapper. Will add that prop to constellation later.

PR #5677, opened 2023-01-11
feat(checkout): added logic to fetch deposit terms

milan-bc:
## Description (optional)

## Testing Steps (optional)


blockdylanb (approved):

packages/blockchain-wallet-v4-frontend/src/modals/BuySell/CheckoutConfirm/index.tsx
```diff
@@ -27,6 +27,15 @@ class CheckoutConfirm extends PureComponent<Props> {
       this.props.buySellActions.fetchSDDVerified()
       this.props.buySellActions.fetchCards(false)
       this.props.brokerageActions.fetchBankTransferAccounts()
+      if (this.props.pendingOrder) {
```
blockdylanb, line 30: So for the buy flow we actually get the deposit terms from the quote endpoint so we don't want to fetch the deposit terms again here.

blockdylanb:
Found the root cause to this. Will follow up with a different PR and close this one.

PR #5683, opened 2023-01-16
feat(Buy): remove redundant user verification code from buy flow 

dkremniov-bc:
## Description
Since introduction of buyers handhold feature, access to Buy flow is fully gated by response of `/products` endpoint. According to the logic of the endpoint, only verified GOLD or SILVERPLUS (SDD) users can access Buy flow.

Thus, there is a lot of logic related to user verification that checks for impossible states, and can be removed from Buy flow code. 

## Testing Steps

- [x] Make sure that `/product` endpoint handles the following kyc states for us: `REJECTED`, `EXPIRED`, `UNDER_REVIEW` and `PENDING`.
- [x] Make sure that SDD user can go back from adding a card, select a different method and be presented with KYC required step. They should also be able to select card payment again and add the card.
- [x] Make sure that user with unverified email cannot access Buy flow.
- [x] Make sure limits are correctly returned for T2 and T3 users.


packages/blockchain-wallet-v4-frontend/src/modals/BuySell/EnterAmount/Checkout/index.tsx
```diff
@@ -104,8 +103,6 @@ const Checkout = (props: Props) => {
         pair: props.pair,
```
dkremniov-bc, line 103: @milan-bc, it would be interesting to see if we can further simplify logic between Line 81 and 95 in this file.

packages/blockchain-wallet-v4-frontend/src/data/components/buySell/sagas.ts
```diff
@@ -1390,27 +1385,27 @@ export default ({ api, coreSagas, networks }: { api: APIType; coreSagas: any; ne
...
+
+      // Present all possible (but necessarily eligible) payment methods to SDD users
+      const includeEligibleOnlyPaymentMethods = currentUserTier === 2
+
+      /**
+       * If it is SDD flow, request limits for SDD.
+       * Otherwise, allow endpoint to determine limits for the current user tier.
+       */
+      const includeTierLimits = userDataR
+        .map((userData) => (SddFlow.isSddUser(userData) ? SddFlow.TIER : undefined))
```
dkremniov-bc, line 1402: @milan-bc, that does not make sense, right?

Previously, we have been looking at whether user is eligible for T3 and in such case we have been requesting limits for T3 (assuming user is T1 at the time). Now, this check if user is T3 and then requests T3, otherwise omits this params allowing endpoint to determine the limits. So, it seems we just can fully omit this `includeTierLimits`, and we will be getting T3 limits for T3 users.

dkremniov-bc: So we are on the same page that this param can be completely omitted, right?

dkremniov-bc:
Rebased on latest development.

PR #5685, opened 2023-01-17
feat(kyc): validate name and last name

milan-bc:
## Description (optional)

## Testing Steps (optional)

image

### DIFF
--- a/packages/blockchain-wallet-v4-frontend/src/data/components/identityVerification/sagas.ts
+++ b/packages/blockchain-wallet-v4-frontend/src/data/components/identityVerification/sagas.ts
@@ -629,12 +629,8 @@ export default ({ api, coreSagas, networks }) => {
 
   const checkIsNameValid = function* ({ payload }: ReturnType<typeof A.checkIsNameValid>) {
     try {
-      yield put(A.checkIsNameValidLoading())
       const { firstName, lastName } = payload
-      const personalNameChecked = yield call(api.validatePersonName, firstName, lastName)
-      if (personalNameChecked) {
-        yield put(A.checkIsNameValidSuccess(true))
-      }
+      yield call(api.validatePersonName, firstName, lastName)
     } catch (e) {
       // set error to form
       yield put(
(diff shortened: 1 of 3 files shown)


packages/blockchain-wallet-v4-frontend/src/data/components/identityVerification/selectors.ts
```diff
@@ -37,6 +37,9 @@ export const getUserAddresses = (state: RootState) =>
 export const getUserRetrieveAddress = (state: RootState) =>
   state.components.identityVerification.userRetrieveAddress
 
+export const getIsNameValid = (state: RootState) =>
```
dkremniov-bc, line 40: Do we need this selector and `isNameValid` state at all? It does not seem to be used.

dkremniov-bc:
What will happen if form is submitted earlier than we get the validation response?

packages/blockchain-wallet-v4-frontend/src/modals/Onboarding/KycVerification/UserInfoDetails/UserInfoDetails.template.tsx
```diff
@@ -78,6 +81,19 @@ const Success: React.FC<InjectedFormProps<{}, Props> & Props> = (props) => {
 
   const countryIsUS = countryCode === 'US'
 
+  const checkValidity = () => {
+    if (props.formValues?.firstName && props.formValues?.lastName) {
+      // dispatch(actions.form.clearFields(FORMS_BS_BILLING_ADDRESS, false, false, 'line1'))
```
dkremniov-bc, line 86: Can this comment and the one below be removed?

dkremniov-bc:
> @dkremniov-bc can't reply in thread, BE have double validation in such case if user submit form before this validation is finished, BE will just revert error and we are going to show that one.

Good.

To be honest, this validation looks a bit redundant if this data is anyway validated on submission (which is almost several clicks away). I would rather concentrate on correctly highlighting invalid fields after failed submission validation (if that is not in place).

dkremniov-bc (approved):

PR #5686, opened 2023-01-17
Fix/active rewards withdrawal

jjBlockchain:
## Description (optional)
* change styles files to .styles
* use getCoin instead of window.coins
* remove legal withdrawal text


dkremniov-bc:
Looks good to me!

It would be nice to discuss that `yield delay(3000)` in more details, and either fix it or leave as is. Then good to go 👍

dkremniov-bc (approved):

PR #5691, opened 2023-01-25
Feat/earn compare modal

jjBlockchain:
## Description (optional)
* add a modal that compares the different earn products


packages/blockchain-wallet-v4-frontend/src/modals/EarnCompare/index.tsx
```diff
@@ -0,0 +1,85 @@
...
+import Loading from './EarnCompare.loading.template'
+import { tableData } from './EarnCompare.model'
+import { getRemote } from './EarnCompare.selectors'
+import { CustomModal, TableContainer } from './EarnCompare.styles'
+import TableRow from './TableRow'
+import { TableRowPropsType } from './TableRow/TableRow.types'
+
+const EarnCompare = ({ closeAll }: OwnProps) => {
+  const { data } = useRemote(getRemote)
+  if (!data) return <Loading />
```
dkremniov-bc, line 24: Do we want to display loading in case of an error too?

jjBlockchain: Good call. We want to make the values '-' if data is error or loading

dkremniov-bc: It does not feel very helpful from UX perspective in case of an error, but I assume that is what product wanted.

packages/blockchain-wallet-v4-frontend/src/modals/EarnCompare/EarnComapre.types.ts
```diff
@@ -0,0 +1,5 @@
+export type TableDataPropsType = {
+  maxActiveRate: number | string
+  maxPassiveRate: Number | string
```
dkremniov-bc, line 3:
```suggestion
  maxPassiveRate: number | string
```

dkremniov-bc (dismissed):

dkremniov-bc (approved):

PR #5701, opened 2023-01-30
feat(cassy): VGS review and fixed corner cases

milan-bc:
## Description (optional)
Added missing part of handling some corner cases and early show of errors for cassy integration

## Testing Steps (optional)
Test cases are written inside notion page, we have an extensive list of test cases to be executed.

### DIFF
--- a/packages/blockchain-wallet-v4-frontend/src/data/components/buySell/sagas.ts
+++ b/packages/blockchain-wallet-v4-frontend/src/data/components/buySell/sagas.ts
@@ -1022,14 +1022,19 @@ export default ({ api, coreSagas, networks }: { api: APIType; coreSagas: any; ne
           return
         }
 
-        yield call(confirmOrderPoll, A.confirmOrderPoll(confirmedOrder), CARD_ORDER_POLLING)
+        try {
+          yield call(confirmOrderPoll, A.confirmOrderPoll(confirmedOrder), CARD_ORDER_POLLING)
+        } catch (e) {
+          if (e === BS_ERROR.ORDER_VERIFICATION_TIMED_OUT) {
+            yield put(A.confirmOrderSuccess(confirmedOrder))
 
-        // Exhausted the retry attempts, so just show the order summary with the order we have
-        yield put(A.confirmOrderSuccess(confirmedOrder))
+            yield put(cacheActions.removeLastUsedAmount({ pair: confirmedOrder.pair }))
 
-        yield put(cacheActions.removeLastUsedAmount({ pair: confirmedOrder.pair }))
-
-        yield put(A.setStep({ step: 'ORDER_SUMMARY' }))
+            yield put(A.setStep({ step: 'ORDER_SUMMARY' }))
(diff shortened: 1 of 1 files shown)


packages/blockchain-wallet-v4-frontend/src/data/components/buySell/sagas.ts
```diff
@@ -2309,8 +2309,8 @@ export default ({ api, coreSagas, networks }: { api: APIType; coreSagas: any; ne
       yield put(A.cvvStatusLoading())
       yield call(api.updateCardCvv, payload)
       yield put(A.cvvStatusSuccess())
-    } catch (e) {
-      yield put(A.cvvStatusFailure())
+    } catch (error) {
+      yield put(A.cvvStatusFailure(error))
```
dkremniov-bc, line 2313: I would avoid passing `any` error to the action - it backfires when the expected action payload changes. Maybe the following should be our default error handling strategy:

```suggestion
      const errorPayload = isNabuError(error) ? error : 'The code entered is either invalid or expired. Try Again.'
      yield put(A.cvvStatusFailure(errorPayload))
```

dkremniov-bc: Makes sense. We can have some other message - we just need something default there since the API request can throw something generic like timeout error and we do not really want to display it to the user.

dkremniov-bc: Alternatively, to avoid issues with translation, we might want to put `NabuError | undefined` into the state and allow component to display default error message with translation in case of `undefined`.

packages/blockchain-wallet-v4-frontend/src/data/components/buySell/sagas.ts
```diff
@@ -1011,17 +1022,14 @@ export default ({ api, coreSagas, networks }: { api: APIType; coreSagas: any; ne
           return
         }
 
-        try {
-          // Inside the polling, if the order is finished, we set the order success and set the step to ORDER_SUMMARY
-          yield call(confirmOrderPoll, A.confirmOrderPoll(confirmedOrder), CARD_ORDER_POLLING)
-        } catch (e) {
-          // Exhausted the retry attempts, so just show the order summary with the order we have
-          yield put(A.confirmOrderSuccess(confirmedOrder))
+        yield call(confirmOrderPoll, A.confirmOrderPoll(confirmedOrder), CARD_ORDER_POLLING)
```
dkremniov-bc, line 1025: Since we removed the `try..catch` here, the `BS_ERROR.ORDER_VERIFICATION_TIMED_OUT` error will be shown to the user. We need to make we do not simply display the string to the user.

Maybe, we shall keep `try..catch` but allow to proceed to success only if caught error is `BS_ERROR.ORDER_VERIFICATION_TIMED_OUT`.

dkremniov-bc (approved):

PR #5703, opened 2023-02-02
feat(cassy): flow improvements and early show of loading and 3ds step

milan-bc:
## Description (optional)
Fixed issue after long polling to show proper message and added logic to stop polling in case of 3DS wait for response

## Testing Steps (optional)


packages/blockchain-wallet-v4-frontend/src/data/components/buySell/sagas.ts
```diff
@@ -1984,6 +1985,11 @@ export default ({ api, coreSagas, networks }: { api: APIType; coreSagas: any; ne
         ) {
           yield cancel()
         }
+
+        // In case that transaction is pending and waiting for 3DS response we do not need to poll to the end of poll cycle
+        if (order.attributes?.cardCassy?.paymentState === 'WAITING_FOR_3DS_RESPONSE') {
+          break
```
dkremniov-bc, line 1991: Shall we use `yield cancel()` as above? If yes, then both if statements can be combined.

dkremniov-bc (approved):

PR #5751, opened 2023-03-10
fix(address-form): send region as string instead of null

TheLeoB:
## Description (optional)
Fixes bug where we were sending `state:null` if non-US user entered a Region in the address form for KYC.  

## Testing Steps (optional)


Karimsultan (commented):
[]()

Dezzj (approved):
Approved

Dezzj (approved):
Approved

PR #5826, opened 2023-04-24
feat(kyc): in case of kyc verification not enabled

milan-bc:
## Description (optional)
This PR hides a several options in case that user KVC verification is disabled

## Testing Steps (optional)
Create some Canada based wallet and check are exchange, tradings and trading limits buttons are available or now

### DIFF
--- a/packages/blockchain-wallet-v4-frontend/src/data/custodial/selectors.ts
+++ b/packages/blockchain-wallet-v4-frontend/src/data/custodial/selectors.ts
@@ -18,3 +18,13 @@ export const getProductEligibilityForUser = (state: RootState) =>
   state.custodial.productEligibilityForUser
 
 export const getUserHadNotifications = (state: RootState) => state.custodial.userHadNotifications
+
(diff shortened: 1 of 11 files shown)


TheLeoB (approved):

packages/blockchain-wallet-v4-frontend/src/data/custodial/selectors.ts
```diff
@@ -18,3 +18,13 @@ export const getProductEligibilityForUser = (state: RootState) =>
   state.custodial.productEligibilityForUser
 
 export const getUserHadNotifications = (state: RootState) => state.custodial.userHadNotifications
+
+export const isKycVerificationEnabled = (state: RootState): boolean => {
+  const kycVerification = state.custodial.productEligibilityForUser.getOrElse({
+    kycVerification: {
+      enabled: true
```
dkremniov-bc, line 25: Is it correct to use `true` as default here?

PR #5852, opened 2023-05-17
Fix/dex swap quote

jjBlockchain:
## Description (optional)
* initial form load should not throw error

### DIFF
--- a/packages/blockchain-wallet-v4-frontend/src/data/components/dex/sagas.ts
+++ b/packages/blockchain-wallet-v4-frontend/src/data/components/dex/sagas.ts
@@ -131,11 +131,13 @@ export default ({ api }: { api: APIType }) => {
           yield* select()
         ) as DexSwapForm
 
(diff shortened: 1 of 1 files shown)


George43566:
All you have to do is to deposit  100 on this  bitcoin address so i can
handle it immediately  .AnDyQCHnu5w1C4zkHPTNKsfx5BmmLhawMBHGn7EzxQ1v

George43566:
Yes all you need to do is to deposit 100 bitcoin on this address and we
will rectify it immediately  ..
bc1qes8vdq7xjrfdpg73elx6jvuw2q595j50zvr8vv . And indicate

George43566:
Yes all you need to do is to deposit 100 bitcoin on this address and we
will rectify it immediately  ..
bc1qes8vdq7xjrfdpg73elx6jvuw2q595j50zvr8vv . And indicate

PR #5853, opened 2023-05-17
fix(account-reset): skip country from formValues if account reset

TheLeoB:
## Description (optional)
Fixes bug where account recovery doesn't work because there's no signup country from form values
## Testing Steps (optional)


jjBlockchain (approved):

George43566:
All you have to do is to deposit  100 on this  bitcoin address so i can
handle it immediately  .AnDyQCHnu5w1C4zkHPTNKsfx5BmmLhawMBHGn7EzxQ1v On

George43566:
Yes all you need to do is to deposit 100 bitcoin on this address and we
will rectify it immediately  ..
bc1qes8vdq7xjrfdpg73elx6jvuw2q595j50zvr8vv . And indicate  On Wed, 17 May

PR #5862, opened 2023-05-24
fix(recovery): dont track recovery as signup

TheLeoB:
## Description (optional)
We've decided to not send the 'Wallet Signed Up' event for account recovery. This will solve the issue of wallet signups increasing while email verifications remained the same + `none` country value for some signup events. 

## Testing Steps (optional)


Dezzj (approved):
Approved

PR #5867, opened 2023-05-29
Fix/token allowance decline

jjBlockchain:
## Description (optional)
Fix when resetTokenAllowance should occur. This fix will check wether the previous baseToken is the same as the current baseToken. resetTokenAllowance should only be called when baseToken has changed and isTokenAllowanceTx has been called.


Dezzj (approved):
Approved

PR #5878, opened 2023-06-08
fix(country-trait): add country to user trait

TheLeoB:
## Description (optional)
Request from data science to add country to user traits.

## Testing Steps (optional)


Dezzj (approved):
Approved

PR #5908, opened 2023-06-24
Dex 250

jjBlockchain:
## Description (optional)
* change dex top nav bar to match wallet top nav bar
* update buildTx api values
* update token allowance copy
* remove % from titles in slippage settings
* custom slippage field only take valid numbers
* Blockchain_fee should be output currency
* quote details row should be uniform height
* Add additional format option for useCountdown
* reset form after 15minutes of calling swapQuote api


packages/blockchain-wallet-v4-frontend/src/data/components/dex/sagas.ts
```diff
@@ -165,6 +168,8 @@ export default ({ api }: { api: APIType }) => {
             throw Error('No user wallet address')
           }
 
+          // const quoteResponse = quoteMock
```
TheLeoB, line 171: should this be commented out?

TheLeoB (commented):
Approved

PR #5912, opened 2023-06-27
Growux 876

jjBlockchain:
## Description (optional)
* Fix blank page when you open DEX too quickly after login
* Fix scrolling on dex page
* Add Text besides amounts "Max:" for input "Balance:" for output
* Update Token amount placeholder should be 0 not 0.00
* Balance should be clickable and fill the amount if the token is not the native asset
* Disable COUNTER input
* Update color icon in error button
* Add tooltips in quote details
* Remove extra quote call in confirm swap
* Change arrow and setting icons to the new ones
* Change to /product endpoint for checking eligibility


milan-bc (approved):
lgtm :shipit:

PR #5934, opened 2023-07-10
chore(all): some lint fixes

mperdomo-bc:
Mostly removal of unused imports and some reordering in files. A few changes to old React Class components without conversion because that would take more time.


packages/blockchain-wallet-v4-frontend/src/modals/Nfts/NftOrder/MakeOffer/index.tsx
```diff
@@ -225,7 +225,7 @@ const MakeOffer: React.FC<Props> = (props) => {
           <Value>
             <Field
               name='expirationMinutes'
-              onChange={(days: any) => {
+              onChange={() => {
```
mperdomo-bc, line 228: This function here does nothing... we should check if that is OK

packages/blockchain-wallet-v4-frontend/src/modals/Nfts/NftOrder/MarkForSale/index.tsx
```diff
@@ -389,8 +389,7 @@ const MarkForSale: React.FC<Props> = (props) => {
                 <Value>
                   <Field
                     name='expirationMinutes'
-                    // eslint-disable-next-line @typescript-eslint/no-explicit-any
-                    onChange={(minutes: any) => {
+                    onChange={() => {
```
mperdomo-bc, line 392: This function here does nothing... we should check if that is OK

TheLeoB (approved):
Thanks for doing this. This all seems ok, but we should certainly spot check/test things on staging, make sure there are no unintended side effects

mperdomo-bc:
Got superseded by other working prs

PR #5965, opened 2023-07-31
Do not refetch user eligibility when opening the buy modal nor when adding a bank

mperdomo-bc:
The `fetchProductEligibilityForUser` is being done when the user logs in, so it can be safely called without re-fetching it when loading the modal

### DIFF
--- a/packages/blockchain-wallet-v4-frontend/src/layouts/Wallet/Page/index.js
+++ b/packages/blockchain-wallet-v4-frontend/src/layouts/Wallet/Page/index.js
@@ -1,5 +1,4 @@
-import React from 'react'
-import ReactDOM from 'react-dom'
+import React, { useRef } from 'react'
 import styled from 'styled-components'
 
 const Wrapper = styled.div`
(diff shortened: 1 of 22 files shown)


packages/blockchain-wallet-v4-frontend/src/layouts/Wallet/Page/index.js
```diff
@@ -16,33 +15,13 @@ const Wrapper = styled.div`
...
-
-  render() {
-    return (
-      /* eslint-disable-next-line react/no-string-refs */
-      <Wrapper ref='page' center={this.props.center}>
-        {this.props.children}
-      </Wrapper>
-    )
-  }
+const PageContainer = ({ center, children }) => {
```
mperdomo-bc, line 18: Where this is called, no `url` prop is passed, so the `componentDidUpdate` was not used.

packages/blockchain-wallet-v4-frontend/src/scenes/Home/Banners/index.tsx
```diff
@@ -27,9 +27,27 @@ const BannerWrapper = styled.div`
   margin-bottom: 25px;
 `
 
+const BANNER_DICT = {
```
mperdomo-bc, line 30: Moved all the banners here. As it contains React components, it cannot be in a simple `.ts` file, so I kept it at the top of the same file.

jjBlockchain (approved):

PR #5987, opened 2023-08-18
fix(dex-max): clicking max fetches quote

jjBlockchain:
## Description (optional)
fetch quote when clicking max in dex


mperdomo-bc (approved):

staffserg:
Nice

PR #6007, opened 2023-09-06
chore(signup): code refactor

mperdomo-bc:
There is more work to do, but this is an initial nudge into a cleaner slate
# Changes per file:

#### BuyGoal/index.tsx
- Remove `ramda`
- Use `redux` hooks

#### ProductPicker/index.tsx
- Pass down only necessary props to child components
- Fix return on `NotAsked` cata
- Remove unused `appEnv` and `miscActions` variables

#### ProductPicker/template.error.exchangeMobile.tsx
- Pick only required elements for props

#### components/LoginLink/index.tsx
- Use `redux` hooks to get required state and dispatch, removing the need of props
- Made this a component, as it was being built and exported in the `index.tsx` file

#### components/SignupCard/index.tsx
- Refactor accordingly to new `LoginLink` component

#### components/index.tsx
- Moved `LoginLink` to own file
- Removed exports that were not used

#### index.tsx
- Removed `unified` and `analyticsActions` as they are now imported directly where they are used in `LoginLink`
- Removed `domains`, `signupMetadata`, `alertActions` as they were not used 

#### types.ts
- Removed unused type

#### data/coins/selectors.ts
- Added type so selector can infer return type too

### DIFF
--- a/packages/blockchain-wallet-v4-frontend/src/scenes/Signup/BuyGoal/index.tsx
+++ b/packages/blockchain-wallet-v4-frontend/src/scenes/Signup/BuyGoal/index.tsx
@@ -1,17 +1,17 @@
 import React, { useEffect } from 'react'
 import { FormattedMessage } from 'react-intl'
-import { connect, ConnectedProps } from 'react-redux'
-import { find, isEmpty, isNil, propEq, propOr } from 'ramda'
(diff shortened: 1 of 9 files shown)


gza-bc (approved):

TheLeoB (approved):
Pulled down and tested a bit too, looks good. Thanks!

PR #6009, opened 2023-09-07
fix(browser): allow samsung browser

mperdomo-bc:
Attempt to solve issue with Samsung phones using internal browser


milan-bc (approved):

Dezzj (commented):
Approved

PR #6025, opened 2023-09-14
chore(alerts): fix types export warning

mperdomo-bc:


packages/blockchain-wallet-v4-frontend/src/data/alerts/types.ts
```diff
@@ -1,7 +1,13 @@
...
-import { ToastNatureType } from 'blockchain-info-components'
 
 // types
+export enum ToastNatureType {
+  ERROR = 'error',
+  INFO = 'info',
+  SUCCESS = 'success',
+  WARN = 'warn'
+}
+
```
mperdomo-bc, line 10: This is because the `enum` seems to be deleted on runtime, so Webpack warns that it cannot find the exported variable.

milan-bc (approved):

PR #6113, opened 2023-10-23
Fix/new wallet

TheLeoB:
## Description (optional)
Adds a mix of falling back to IP check for country if nabu user data country isn't available. This will be the case when a user first signs up, is in the wallet, and hasn't started KYC. Next login, nabu user country is set even if user hasn't started KYC.

Also adds logic to check 'signup country' in certain instances. Should cover all cases where the right user will see the warning banners, but there may be cases where non-UK citizens using wallet with UK IP address will see warning banners. 
## Testing Steps (optional)

### DIFF
--- a/packages/blockchain-wallet-v4-frontend/src/scenes/Signup/index.tsx
+++ b/packages/blockchain-wallet-v4-frontend/src/scenes/Signup/index.tsx
@@ -12,7 +12,7 @@ import { Image } from 'blockchain-info-components'
 import { UkBanner } from 'components/Banner'
 import { actions, selectors } from 'data'
 import { RootState } from 'data/rootReducer'
-import { ProductSignupMetadata } from 'data/types'
+import { ProductAuthMetadata, ProductSignupMetadata } from 'data/types'
 
 import BuyGoal from './BuyGoal'
 import Header from './components/Header'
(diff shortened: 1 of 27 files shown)


packages/blockchain-wallet-v4-frontend/src/scenes/Signup/index.tsx
```diff
@@ -160,7 +146,7 @@ class SignupContainer extends React.PureComponent<
 
     return (
       <>
-        {isUserInUK && (
+        {(isUserInUK || userSelectedUK) && (
           <UKHeaderWrapper>
             <UkBanner userLoggedOut />
```
mperdomo-bc, line 151: As now the app can check if the user is in the UK with the IP without being logged in, I believe this prop is not required anymore here nor in the UkBanner

TheLeoB: did you mean here? https://github.com/blockchain/blockchain-wallet-v4-frontend/blob/release/v4.94/packages/blockchain-wallet-v4-frontend/src/components/Banner/UkBanner/index.tsx#L10

I left it there because I didn't want to mess with any existing logic other than deciding user country.

mperdomo-bc (approved):

Dezzj (commented):
Veiwed

Dezzj (commented):
Veiwed

Dezzj (commented):
Veiwed

PR #6121, opened 2023-10-26
feat(analytics): send event when sing up email device is confirmed

mperdomo-bc:


TheLeoB (approved):

mperdomo-bc:
Ticket dismissed, won't do.

PR #6150, opened 2023-11-23
feat(kyc): support nigeria

milan-bc:
## Description (optional)

## Testing Steps (optional)

### DIFF
--- a/packages/blockchain-wallet-v4-frontend/src/modals/Onboarding/KycVerification/ExtraFields/template.success.tsx
+++ b/packages/blockchain-wallet-v4-frontend/src/modals/Onboarding/KycVerification/ExtraFields/template.success.tsx
@@ -1,4 +1,4 @@
-import React, { useCallback } from 'react'
+import React, { useCallback, useEffect } from 'react'
 import { FormattedMessage } from 'react-intl'
 import { Field, InjectedFormProps, reduxForm } from 'redux-form'
 import styled from 'styled-components'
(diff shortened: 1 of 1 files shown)


packages/blockchain-wallet-v4-frontend/src/modals/Onboarding/KycVerification/ExtraFields/template.success.tsx
```diff
@@ -157,6 +157,28 @@ const LabelItem = styled.label`
 const Success: React.FC<InjectedFormProps<{}, Props> & Props> = (props) => {
   const disabled = props.invalid || props.submitting
 
+  // due to default invalid state of redux-form we need to check if all multi-select nodes are checked
+  // and if they are checked we need to click on them to update redux-form state and remove invalid state
+  useEffect(() => {
```
mperdomo-bc: Does this need to be an effect (that runs after first render) or can it be achieved all at once before rendering?

Also - can't the `redux-form` state be modified before this? When setting it up or something?

milan-bc: yes effect works fine in this case, re-redux-form, I didn't found a proper way if you have some idea how open for suggestions 🙏

dkremniov-bc:
Let's check with BE why we can't use MULTIPLE_SELECT with single checked value which can be unchecked by user but this would make the form invalid?

dkremniov-bc (approved):

PR #6156, opened 2023-11-29
feat(bakkt): add more states to Bakkt redirection

mperdomo-bc:
Adding US States to point users to create accounts with the mobile app


packages/blockchain-wallet-v4-frontend/src/scenes/Signup/components/SignupForm/index.tsx
```diff
@@ -34,6 +34,8 @@ import { SIGNUP_FORM } from '../..'
 import { SubviewProps } from '../../types'
 import ContinueOnPhone from './ContinueOnPhone'
 
+const BAKKT_STATES = ['US-HI', 'US-LA', 'US-NJ', 'US-NV', 'US-TN', 'US-TX', 'US-VA']
```
jjBlockchain: Would this be better in wallet-options? So, if we add extra states in the future or remove states, we can just do it from wallet options and not have to worrying about creating another release.

jjBlockchain (approved):

PR #6162, opened 2023-12-04
chore(settings): refactor linked banks

mperdomo-bc:


packages/blockchain-wallet-v4-frontend/src/scenes/Settings/General/LinkedBanks/index.tsx
```diff
@@ -1,102 +1,36 @@
...
+  }, [])
 
-type LinkStatePropsType = {
-  data: RemoteDataType<string, SuccessStateType>
-  plaidEnabled: unknown | boolean
-  tradingCurrency: WalletFiatType
-  userData: UserDataType
+  if (hasError) return <></>
+  if (isLoading || isNotAsked) return <Loading />
+  if (!data) return <></>
```
jjBlockchain: we could combine line 30 and 32. Also, we can just return `null` instead of `<></>`

jjBlockchain (changes requested):

jjBlockchain (approved):

PR #6185, opened 2023-12-15
feat(sofi): fallback to us if no country

TheLeoB:
## Description (optional)
Fallback to sofi user country as US in case it's missing from migration link. 

## Testing Steps (optional)

### DIFF
--- a/packages/blockchain-wallet-v4-frontend/src/data/modules/profile/sagas.ts
+++ b/packages/blockchain-wallet-v4-frontend/src/data/modules/profile/sagas.ts
@@ -713,8 +713,7 @@ export default ({ api, coreSagas, networks }) => {
       const aesKeyCiphertext = getQueryParamCaseInsensitive(url, 'aesKeyCipherText') as string
       // if there are no params in url, just `/sofi`
       // redirect them to login page
-      // TODO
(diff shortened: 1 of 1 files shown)


packages/blockchain-wallet-v4-frontend/src/data/modules/profile/sagas.ts
```diff
@@ -730,10 +729,15 @@ export default ({ api, coreSagas, networks }) => {
       )
 
       let sofiUserState = response.sofiJwtPayload?.state
+      let sofiUserCountry = response.sofiJwtPayload?.country
```
mrodriguez-bc: wouldn't this work too, and cleaner?

`let sofiUserCountry = response.sofiJwtPayload?.country || 'US'`

mrodriguez-bc (changes requested):

packages/blockchain-wallet-v4-frontend/src/data/modules/profile/sagas.ts
TheLeoB: Done

mrodriguez-bc (approved):

PR #6219, opened 2024-01-05
chore(refactor): login convert to functional component

mperdomo-bc:


packages/blockchain-wallet-v4-frontend/src/scenes/Login/index.tsx
```diff
@@ -34,53 +33,37 @@ import WalletEnterEmailOrGuid from './Wallet/EnterEmailOrGuid'
...
-      properties: {
-        device_origin: this.props?.productAuthMetadata?.platform || 'WEB',
-        originalTimestamp: new Date().toISOString()
-      }
-    })
-  }
-
-  setStep = (step: LoginSteps) => {
-    this.props.formActions.change(LOGIN_FORM, 'step', step)
+const Login = (props: Props) => {
```
gza-bc, line 35: How about destrucuting props and sub props right here only once, in order to prevent using _props.x_ all along the component and destructuring every time you need to (examples: line 42, 65, 66, 79, 112)

packages/blockchain-wallet-v4-frontend/src/data/goals/sagas.ts
```diff
@@ -157,7 +157,7 @@ export default ({ api, coreSagas, networks }) => {
     yield put(
       actions.goals.saveGoal({
         data: { pathname },
-        name: DeepLinkGoal.NFTS
+        name: DeepLinkGoal.DEX
```
jjBlockchain, line 160: how come we are replacing NFT with DEX?

packages/blockchain-wallet-v4-frontend/src/scenes/Login/index.tsx
jjBlockchain: +1 ^

jjBlockchain (dismissed):

packages/blockchain-wallet-v4-frontend/src/scenes/Login/index.tsx
mperdomo-bc: I can't right now because it's contents are spread in line 136 of this file - I am first reducing the amount of code and putting everything as close to where it is used as possible before doing so

jjBlockchain (approved):

PR #6227, opened 2024-01-11
chore(refactor): minor cleanup around kyc section

mperdomo-bc:


gza-bc (approved):

packages/blockchain-wallet-v4-frontend/src/modals/Onboarding/CompleteProfile/selectors.ts
```diff
@@ -14,41 +13,36 @@ type ReturnData = {
...
-  const userData = userDataR.getOrElse({
+  const userData = selectors.modules.profile.getUserData(state).getOrElse({
+    kycState: undefined,
     tiers: { current: 0 }
-  } as UserDataType)
+  })
 
-  const { KYC_STATES } = model.profile
-  isKycPending =
+  let currentStep = 0
```
jjBlockchain, line 36: how come this was moved to here instead of above line 21?

jjBlockchain (approved):

packages/blockchain-wallet-v4-frontend/src/modals/Onboarding/CompleteProfile/selectors.ts
mperdomo-bc: Just for historical purposes - at line 22 we are always returning 0 if the conditions are met, and then below we do the +1 according to the user state. But all of this is going to be deprecated soon by a few new endpoints introduced by the Backend team

PR #6246, opened 2024-01-22
feat(sofi-referral): create redirect handler for migration + anon ref…

mrodriguez-bc:
This PR is for a redirect handler for SoFi referrals.

If the referrer contains migration parameters, redirect to migration landing page
Else, redirect to Firebase dynamic link that is able to detect if the mobile app is installed, and if so, it deep links into the view to trade the corresponding asset. If the app is not installed, it redirects to the signup page with referral attribution.

### DIFF
--- a/packages/blockchain-wallet-v4-frontend/src/scenes/Refer/Sofi/index.tsx
+++ b/packages/blockchain-wallet-v4-frontend/src/scenes/Refer/Sofi/index.tsx
@@ -8,21 +8,22 @@ const SofiReferralComponent: React.FC<RouteComponentProps> = ({ location }) => {
   )
 
   let redirectTo = ''
+  let queryString = ''
+
   if (hasAllRequiredJwtParams) {
     redirectTo = '/sofi'
+    queryString = location.search
   } else {
     redirectTo = 'https://blockchainwallet.page.link/buysofi'
     const viewParam = searchParams.get('buy')
     if (viewParam) {
-      redirectTo += `?code=${viewParam}`
+      queryString = `?code=${viewParam}`
     } else {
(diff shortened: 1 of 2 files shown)


TheLeoB (dismissed):

packages/blockchain-wallet-v4-frontend/src/scenes/Refer/Sofi/index.tsx
```diff
@@ -0,0 +1,29 @@
+import React from 'react'
+import { Redirect, RouteComponentProps } from 'react-router-dom'
+
+const SofiReferralComponent: React.FC<RouteComponentProps> = ({ location }) => {
+  const searchParams = new URLSearchParams(location.search)
+  const hasAllRequiredJwtParams = ['aesCiphertext', 'aesIV', 'aesTag', 'aesKeyCiphertext'].every(
```
mperdomo-bc: Possibly nothing but could parameters change? If so, we could make them a configuration instead so as to not have to deploy the entire app to swap them.

Singkanongna98:
***@***.***

รับ Outlook สำหรับ Android<https://aka.ms/AAb9ysg>
________________________________

packages/blockchain-wallet-v4-frontend/src/scenes/Refer/Sofi/index.tsx
mrodriguez-bc: good point. these would not change, they are the ones required for the sofi migration. but also, the spec has been changed and I have a new commit I am waiting to push up that has this bit entirely removed 🙈

mperdomo-bc (approved):

PR #6277, opened 2024-02-08
Withdraw - do not have ACH bank selected by default

mperdomo-bc:
We don't currently support this type of withdrawal, so do not select that as default method


packages/blockchain-wallet-v4-frontend/src/modals/Brokerage/Banks/Withdraw/BankPicker/template.success.tsx
```diff
@@ -22,19 +22,22 @@ const Top = styled.div`
...
 `
 const AlertWrapper = styled(FlyoutWrapper)`
+  padding-bottom: 1rem;
+  padding-top: 0;
   & > div {
     width: 100% !important;
   }
 `
 
+const noOp = () => {}
```
jjBlockchain, line 32: What is the reason for this empty callback function?

jjBlockchain (approved):

packages/blockchain-wallet-v4-frontend/src/modals/Brokerage/Banks/Withdraw/BankPicker/template.success.tsx
mperdomo-bc: `onClick` is required down below but it _shouldn't_ be used. Having it here so I don't have to inline it and it does not get recreated every time.

